PR #123, opened 2023-12-11
Docs-add-flare-notebook

mendonk:
* Add FLARE notebook to index, docs page, nav, and colab link (not published yet). 
* Added a [partial](https://github.com/datastax/ragstack-ai/compare/docs-add-flare-notebook?expand=1#diff-83f94f51e2ef5cbee0ea79897ab5e7a9e2b00147d3eaf1ab99c26a94b41062db) for ragstack and astra setup -  if we can standardize on these setup variables, it will make setup easier to read and require less writing.

WDYT @jordanrfrazier ?

astra_token = os.getenv("ASTRA_DB_APPLICATION_TOKEN")
astra_endpoint = os.getenv("ASTRA_DB_API_ENDPOINT")
collection = os.getenv("ASTRA_DB_COLLECTION")
openai_key = os.getenv("OPENAI_API_KEY")

### DIFF
--- a/docs/modules/examples/pages/flare.adoc
+++ b/docs/modules/examples/pages/flare.adoc
@@ -1,0 +1,119 @@
+= Foward-Looking Augmented REtrieval (FLARE)
+
+image::https://colab.research.google.com/assets/colab-badge.svg[align="left",link="https://colab.research.google.com/github/datastax/ragstack-ai-examples/blob/main/FLARE.ipynb"]
+
+FLARE is an advanced retrieval technique that combines retrieval and generation in LLMs.
+It enhances the accuracy of responses by iteratively predicting the upcoming sentence to anticipate future content when the model encounters a token it is uncertain about.
+See the original repository link:https://github.com/jzbjyb/FLARE/tree/main[here]{external-link-icon} for more detail.
+
+The basic workflow is:
+
+. Send a query.
+. The model generates tokens while iteratively predicting the upcoming sentence.
+. If the model sees a token with a low probability level, it uses the predicted sentence as a query to retrieve new, relevant documents.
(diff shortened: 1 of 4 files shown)


docs/modules/examples/pages/flare.adoc
```diff
@@ -0,0 +1,119 @@
...
+
+FLARE is an advanced retrieval technique that combines retrieval and generation in LLMs.
+It enhances the accuracy of responses by iteratively predicting the upcoming sentence to anticipate future content when the model encounters a token it is uncertain about.
+See the original repository link:https://github.com/jzbjyb/FLARE/tree/main[here]{external-link-icon} for more detail.
+
+The basic workflow is:
+
+. Send a query.
+. The model generates tokens while iteratively predicting the upcoming sentence.
+. If the model sees a token with a low probability level, it uses the predicted sentence as a query to retrieve new, relevant documents.
```
kerinin, line 13: Can we change "low probability" to "low confidence"? Technically, at every generation step the probabilities of each token and taking are computed and the most likely token (modulo temperature) is returned. FLARE looks to see if the token probabilities have low entropy (ie, all the probabilities are the same) indicating low confidence in any particular token. In practice, FLARE estimates the entropy using the max probability, since the max probability is minimized when confidence is minimized - so "low probability" isn't _wrong_ per se, just less accurate than "low confidence".

kerinin (approved):

docs/modules/examples/pages/flare.adoc
```diff
@@ -0,0 +1,119 @@
+= Foward-Looking Augmented REtrieval (FLARE)
```
bjchambers, line 1: s/Foward/Forward

PR #128, opened 2023-12-11
[RAG-75] Docs: add FLARE notebook

jordanrfrazier:
Uses `nbmake` to delete the collection at the end of this notebook. This will be standard across all notebook tests so we don't a) use the same collection through multiple tests and b) hit the collection limit in astra


epinzur (approved):

examples/notebooks/FLARE.ipynb
```diff
@@ -0,0 +1,668 @@
...
+        {
+            "cell_type": "markdown",
+            "metadata": {
+                "nbmake": {
+                    "post_cell_execute": [
+                        "# Delete collection after test so other tests don't re-use or hit limit\n",
+                        "vstore.delete_collection()\n"
+                    ]
+                }
+            },
```
epinzur, line 641: Maybe don't make this a nbmake call, but actually expose it in the notebook so users can cleanup the created collection from their astra.

jordanrfrazier: I'll actually do both -- we want the CI to clean up, and I'll expose it for the users in a Cleanup section

PR #129, opened 2023-12-12
Llama v0.9.14.post1

eolivelli:

PR #131, opened 2023-12-12
Add license to package metadata

nicoloboschi:


ragstack-e2e-tests/e2e_tests/langchain_llamaindex/test_astra.py
```diff
@@ -18,9 +18,6 @@
 from llama_index.llms import OpenAI
 from llama_index.vector_stores import (
     AstraDBVectorStore,
-    MetadataFilters,
```
eolivelli, line 21: This is not related

eolivelli: I added in #129 the fix for the lint problems.

I am still curious about how is it possible that we still commit broken files according to style

eolivelli (approved):

PR #132, opened 2023-12-12
Add explicit dep for langchain-community

nicoloboschi:


eolivelli (approved):
+1

cbornet (changes requested):
We should probably also import the git main dep in `pyproject.langchain.toml`

PR #139, opened 2023-12-13
CI: enable dev for sched jobs

nicoloboschi:


.github/workflows/ci.yml
```diff
@@ -53,11 +53,11 @@ jobs:
             env: PROD
             region: us-east-2
             cloud: aws
-#          - name: Astra DEV
-#            astra_token_secret: E2E_TESTS_ASTRA_DEV_DB_TOKEN
-#            env: DEV
-#            region: us-west-2
-#            cloud: aws
+          - name: Astra DEV
```
nicoloboschi, line 56: will be commented out, just want to check if dev actually works

eolivelli, line 56: we said in slack that it is better to not run PR validation against DEV, it is only a pain.

we should have a scheduled workflow that runs all the tests against DEV every X hours

nicoloboschi: https://github.com/datastax/ragstack-ai/pull/139#discussion_r1425154334

PR #140, opened 2023-12-13
CI: skip tests/notebooks when not needed

nicoloboschi:


.github/workflows/ci.yml
```diff
@@ -19,6 +20,18 @@ concurrency:
...
 jobs:
+  preconditions:
+    name: Preconditions
+    runs-on: ubuntu-latest
+    outputs:
+      notebooks: ${{ steps.filter.outputs.notebooks }}
+      e2e_tests: ${{ steps.filter.outputs.e2e_tests }}
+    steps:
+      - uses: actions/checkout@v3
+      - uses: dorny/paths-filter@v2
```
eolivelli, line 31: what is this ?
dorny ? is it a private user ?
we shouldn't depend on unknown stuff,
maybe we can copy/paste the action ?

nicoloboschi: well, it's a pretty established action: https://github.com/dorny/paths-filter

I don't see any reason for not using it. Also we're not passing to it any secret.

PR #141, opened 2023-12-13
CI: delete Astra DB async

nicoloboschi:


.github/actions/cleanup-astra-db/action.yml
```diff
@@ -20,4 +20,4 @@ runs:
       TERM: linux
     run: |
       (curl -Ls "https://dtsx.io/get-astra-cli" | bash) || true
-      /home/runner/.astra/cli/astra db delete "${{ inputs.db-name }}" --token ${{ inputs.astra-token }} --env "${{ inputs.env }}"
\ No newline at end of file
+      /home/runner/.astra/cli/astra db delete "${{ inputs.db-name }}" --token ${{ inputs.astra-token }} --env "${{ inputs.env }}" --async
```
eolivelli, line 23: what's the difference ?

if it fails we have a leak

currently if it fails then we can restart the clean up

eolivelli (approved):
LGTM

PR #145, opened 2023-12-13
[RAG-40] - Docs - langchain migration

mendonk:
Documentation for LangChain migration.

### DIFF
--- a/docs/modules/ROOT/pages/migration.adoc
+++ b/docs/modules/ROOT/pages/migration.adoc
@@ -1,0 +1,288 @@
+= Migration
+
+To migrate an existing LangChain application to RAGStack and use RAGStack's pinned, tested version of LangChain instead, you need to uninstall LangChain and install RAGStack.
+
+. `pip uninstall langchain`
+. `pip install ragstack-ai --upgrade-strategy="only-if-needed"`
+. `pip list` to see your current list of packages.
+
+An example migration is shown below.
+
+== RAGStack packages
+RAGStack contains the following packages as of version 0.3.1.
+You can replace any of these packages in your existing application with the `ragstack-ai` version.
+
+.RAGStack 0.3.1 packages
+[%collapsible%open]
+====
+[source,console]
+----
+Package             Version
+------------------- ------------
+aiohttp             3.9.1
+aiosignal           1.3.1
+annotated-types     0.6.0
+anyio               4.1.0
+astrapy             0.6.2
+attrs               23.1.0
+backoff             2.2.1
+beautifulsoup4      4.12.2
+cassandra-driver    3.28.0
+cassio              0.1.3
+certifi             2023.11.17
+chardet             5.2.0
+charset-normalizer  3.3.2
+click               8.1.7
+dataclasses-json    0.6.3
+Deprecated          1.2.14
+distro              1.8.0
+emoji               2.9.0
+filetype            1.2.0
+frozenlist          1.4.0
+fsspec              2023.12.2
+geomet              0.2.1.post1
+greenlet            3.0.2
+h11                 0.14.0
+h2                  4.1.0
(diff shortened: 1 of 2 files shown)


docs/modules/ROOT/pages/migration.adoc
```diff
@@ -0,0 +1,288 @@
+= Migration
+
+To migrate an existing LangChain application to RAGStack and use RAGStack's pinned, tested version of LangChain instead, you need to uninstall LangChain and install RAGStack.
```
kerinin, line 3: Could we move this to an aside? I feel like the opening here ought to be something like "Migrating an existing LangChain application to RAGStack is super easy - just change your `requirements.txt` to use `ragstack-ai` and you'll get the benefits of RAGStack, such as <...>". Then we can have an aside that says something like "if you run into problems, you uninstalling your current langchain package could help".

kerinin (approved):

docs/modules/ROOT/pages/migration.adoc
bjchambers: Or pyproject.toml depending on how you created your Python project.

docs/modules/ROOT/pages/migration.adoc
```diff
@@ -0,0 +1,288 @@
...
+
+. `pip uninstall langchain`
+. `pip install ragstack-ai --upgrade-strategy="only-if-needed"`
+. `pip list` to see your current list of packages.
+
+An example migration is shown below.
+
+== RAGStack packages
+RAGStack contains the following packages as of version 0.3.1.
+You can replace any of these packages in your existing application with the `ragstack-ai` version.
```
cbornet, line 13: RAGstack doesn't contain these packages. They are transitive dependencies and their final versions will be resolved on the user environment.
I think we should only mention the dependency versions that we pin (langchain, llama-index, astrapy, cassio, ...)
Currently
```toml
[tool.poetry.dependencies]
python = ">=3.9,<3.12"
astrapy = "~0.6.2"
cassio = "~0.1.3"
unstructured = "^0.10"
llama-index = { version = "0.9.14", extras = ["langchain"] }
langchain = { version = "0.0.343", extras = ["openai"] }
langchain-core = "0.0.8"
```

docs/modules/ROOT/pages/migration.adoc
cbornet: Uninstalling lanchain is not really needed.
Installing RAGStack will replace the langchain version that was there previously if any.

docs/modules/ROOT/pages/migration.adoc
```diff
@@ -0,0 +1,288 @@
...
+. This application requires installation of the following packages:
++
+[source,python]
+----
+pip install langchain datasets openai astrapy tiktoken python-dotenv
+----
++
+. You decide you want to use RAGStack's pinned, tested version of LangChain (`langchain-0.0.349`) instead of the latest version of LangChain (`langchain-0.0.350`).
++
+*You must uninstall LangChain first.*
```
cbornet, line 166: I'm really surprised this is needed.

docs/modules/ROOT/pages/migration.adoc
```diff
@@ -0,0 +1,288 @@
...
+pip uninstall langchain
+Successfully uninstalled langchain-0.0.350
+----
++
+. Install the `ragstack-ai` package with the `--upgrade-strategy="only-if-needed"` option.
+This ensures pip will not upgrade any packages that are already installed, unless required by the `ragstack-ai` package.
++
+[source,python]
+----
+pip install --upgrade-strategy="only-if-needed"
```
cbornet, line 179:
```suggestion
pip install --upgrade-strategy="only-if-needed" ragstack-ai
```

PR #159, opened 2023-12-16
Upgrade llama-index and reactivate Bedrock tests

cbornet:


cbornet:
Draft for now to see if CI is passing.
I will wait for LI 0.9.16 for the final PR.

pyproject.toml
```diff
@@ -14,7 +14,7 @@ python = ">=3.9,<3.12"
 astrapy = "~0.6.2"
 cassio = "~0.1.3"
 unstructured = "^0.10"
-llama-index = { version = "0.9.14", extras = ["langchain"] }
+llama-index = { version = "0.9.15.post2", extras = ["langchain"] }
```
eolivelli, line 17: IIRC we said that we don't want to depend on .post releases. Maybe we can wait for the next train. There is no hurry

pyproject.toml
```diff
@@ -14,7 +14,7 @@ python = ">=3.9,<3.12"
 astrapy = "~0.6.2"
 cassio = "~0.1.3"
 unstructured = "^0.10"
-llama-index = { version = "0.9.14", extras = ["langchain"] }
+llama-index = { version = "0.9.16.post1", extras = ["langchain"] }
```
cbornet, line 17: Had to use `0.9.16.post1` and not `0.9.16` because llama-index uses `post` versions instead of increasing minor to fix bugs 😞 .

eolivelli (approved):
+1

cbornet:
We can't upgrade llama-index because the 0.9.16.post1 release breaks our tests.

PR #171, opened 2023-12-22
Esp/all the datasets

msmygit:


nicoloboschi (changes requested):
please move this stuff under examples/notebooks

PR #175, opened 2024-01-08
Bump LLamaIndex to 0.9.29 and add multimodal Gemini test

nicoloboschi:


ragstack-e2e-tests/e2e_tests/llama_index/test_compatibility_rag.py
```diff
@@ -27,37 +34,120 @@
...
     delete_all_astra_collections,
     delete_astra_collection,
+    AstraRef,
 )
+from llama_index.vector_stores.types import VectorStore, VectorStoreQuery
+from sqlalchemy import Float
+from vertexai.vision_models import MultiModalEmbeddingModel, Image
+
+
+class VectorStoreWrapper:
```
cbornet, line 44: I would call it `VectoreStore` directly.
It will not collide with LI's VectorStore if we remove `as_vector_store `

ragstack-e2e-tests/e2e_tests/llama_index/test_compatibility_rag.py
```diff
@@ -27,37 +34,120 @@
...
+    @abstractmethod
+    def init(self, embedding_dimension: int) -> str:
+        pass
+
+    @abstractmethod
+    def as_vector_store(self) -> VectorStore:
```
cbornet, line 50: We could do an abstract class rather than a pure interface.

ragstack-e2e-tests/e2e_tests/llama_index/test_compatibility_rag.py
```diff
@@ -27,37 +34,120 @@
...
+            self.vector_store.client.vector_find(
+                vector,
+                limit=limit,
+            ),
+        )
 
 
 @pytest.fixture
 def astra_db():
     astra_ref = get_astra_ref()
```
cbornet, line 139: Not directly related to this PR but we should make astra_ref a pytest fixture.

cbornet:
LGTM. Some minor comments.

ragstack-e2e-tests/e2e_tests/llama_index/test_compatibility_rag.py
nicoloboschi: I did it on purpose to being able to catch errors in the actual init method and not in the fixture construction

PR #186, opened 2024-01-11
base docker file

zzzming:


docker/Dockerfile.base
```diff
@@ -0,0 +1,3 @@
+FROM python:3.11
```
nicoloboschi, line 1: what about using a fixed version ?

docker/Dockerfile.multistage
```diff
@@ -0,0 +1,28 @@
...
+
+# The runtime image, used to just run the code provided its virtual environment
+FROM python:3.11-slim-buster as runtime
+
+ENV VIRTUAL_ENV=/app/.venv \
+    PATH="/app/.venv/bin:$PATH"
+
+COPY --from=builder ${VIRTUAL_ENV} ${VIRTUAL_ENV}
+
+# run the app
```
nicoloboschi, line 27: please drop comments

nicoloboschi (changes requested):
@zzzming I don't understand this pr. should this include preinstalled ragstack-ai package?

docker/Dockerfile.multistage
zzzming: fixed

PR #189, opened 2024-01-11
dockerfile example to down model locally part of Docker image

zzzming:


docker/local-llm/Makefile
```diff
@@ -0,0 +1,5 @@
+.PHONY: build_docker
+
+docker:
+	docker build -t ragstack-basic .
```
nicoloboschi, line 4: local-llm ?

zzzming: This is what Ben suggested to support use case that huggingface to download a llm to docker image. Does that make sense?

nicoloboschi: yeah I meant to use it in the docker command 😃 
I updated the code for you

PR #209, opened 2024-01-17
retry on 504 when delete_collection

zzzming:
delete_collection() was in VectorStore class in LC. It probably does NOT make sense to fix their either since this exposes our bug in LC project.
I’m going to add retry delete in the ragstack-ai’s test. I suppose the astrapy repo just throw 504 based on this call raise_for_status()
    def request(self: T) -> API_RESPONSE:
        # Make the raw request to the API
        self.response = self.raw_request()

        # If the response was not successful (non-success error code) raise an error directly
        self.response.raise_for_status()

        # Otherwise, process the successful response
        return self._process_response()

### DIFF
--- a/ragstack-e2e-tests/e2e_tests/langchain/test_astra.py
+++ b/ragstack-e2e-tests/e2e_tests/langchain/test_astra.py
@@ -460,19 +460,14 @@ def environment():
 
 
 def close_vector_db(vector_store: VectorStore):
-    try_close_vector_db(vector_store)
-
-def try_close_vector_db(vector_store: VectorStore, max_attempts=5):
-    for attempt in range(max_attempts):
-        try:
-            vector_store.astra_db.delete_collection(vector_store.collection_name)
-            break
-        except requests.HTTPError as e:
-            if e.response.status_code == 504:
-                logging.error(f"Attempt {attempt+1} 504 error while deleting collection {vector_store.collection_name}: {e}")
-            else:
-                logging.error(f"Error while deleting collection {vector_store.collection_name}: {e}")
-                raise e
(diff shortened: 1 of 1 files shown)


ragstack-e2e-tests/e2e_tests/langchain/test_astra.py
```diff
@@ -2,6 +2,8 @@
 import logging
 from typing import List
 
+import requests
```
nicoloboschi, line 5: no need

ragstack-e2e-tests/e2e_tests/langchain/test_astra.py
```diff
@@ -458,7 +460,19 @@ def environment():
...
 def close_vector_db(vector_store: VectorStore):
-    vector_store.astra_db.delete_collection(vector_store.collection_name)
+    try_close_vector_db(vector_store)
+
+def try_close_vector_db(vector_store: VectorStore, max_attempts=5):
+    for attempt in range(max_attempts):
+        try:
+            vector_store.astra_db.delete_collection(vector_store.collection_name)
+            break
+        except requests.HTTPError as e:
```
nicoloboschi, line 470: please add a sleep in order to relax the call rate

nicoloboschi (changes requested):
Most of the tests call the method [here](https://github.com/datastax/ragstack-ai/blob/main/ragstack-e2e-tests/e2e_tests/conftest.py)

Please adjust the code to put all the collection deletion in one single point.

Also, see inline comments

nicoloboschi:
@zzzming closing this pr since we went ahead with another solution

PR #215, opened 2024-01-19
tests: perform astra collection deletes async

nicoloboschi:


ragstack-e2e-tests/e2e_tests/conftest.py
```diff
@@ -53,34 +61,50 @@ def get_astra_ref() -> AstraRef:
...
+def ensure_astra_env_clean():
+    collections = (
+        DEFAULT_ASTRA_CLIENT.get_collections().get("status").get("collections")
+    )
+    if len(collections) >= 3:
+        logging.info(
+            f"Awaiting for 2 collection slots to be available, currently there are {len(collections)} collections"
+        )
+        delete_all_astra_collections()
+        ensure_astra_env_clean()
```
jordanrfrazier, line 79: Should we add a backoff to here so as to not overwhelm data api with delete timeouts, or perhaps in the `delete_all_astra_collections()`?

jordanrfrazier: Actually curious about this logic -- if deletions are running asynchronously from the tests, isn't it possible that this `delete_all` will delete a collection that we're currently using in a test?

nicoloboschi: you're right, it's better to start a test with 0 collection on the db. 
also, I added a simple sleep between the retries - it should be enough

jordanrfrazier (approved):

PR #227, opened 2024-01-24
initial add of TruLens RAG evaluation framework

epinzur:
This contains the primary scripts I've been using for running medium-scale evaluation of different RAG techniques.

A following PR will contain the results from my first test.


evaluation/tru_shared.py
```diff
@@ -0,0 +1,159 @@
...
+            model="text-embedding-ada-002",
+            api_key=os.getenv("AZURE_OPENAI_API_KEY"),
+            api_version="2023-05-15",
+            temperature=temperature,
+        )
+    else:
+        raise Exception(f"Unknown framework: {framework} specified for getEmbeddingsModel()")
+
+
+def getAstraVectorStore(framework: Framework, collection_name: str):
```
jordanrfrazier, line 142: python convention is `get_astra_vector_store`: https://www.techversantinfotech.com/python-naming-conventions-points-you-should-know/

evaluation/tru_shared.py
```diff
@@ -0,0 +1,159 @@
...
+# OPENAI_API_VERSION
+# TRULENS_DB_CONN_STRING
+
+load_dotenv()
+
+temperature = 0
+
+class Framework(Enum):
+    LANG_CHAIN = "langChain"
+    LLAMA_INDEX = "llamaIndex"
```
jordanrfrazier, line 36: don't technically have to assign values

evaluation/README.md
```diff
@@ -0,0 +1,96 @@
+# Evaluation Metrics
```
jordanrfrazier, line 1: Would add instructions on how to run in readme

evaluation/tru_shared.py
epinzur: updated all method names in `tru_shared.py`

evaluation/README.md
epinzur: I'm going to hold off on doing this until we make a decision if we are going to publicize TruLens or not. For now I'm the only one running these and I know how to do so.

PR #239, opened 2024-01-26
add shareable langsmith links to the test report

jordanrfrazier:
So far only adds the shareable link to the results.xml file. Next step is figuring out how to annotate xml file for testspace.


ragstack-e2e-tests/e2e_tests/langchain/rag_application.py
```diff
@@ -213,10 +230,38 @@ def run_conversational_rag(
...
+    with callbacks.collect_runs() as cb:
+        result = conversation.invoke(
+            {"question": "what is MyFakeProductForTesting?"},
+            config={"callbacks": [tracer]},
+        )
+        run_id = cb.traced_runs[0].id
+        record_langsmith_sharelink(LANGSMITH_CLIENT, run_id, record_property)
+        logging.info("First result: " + str(result))
+
+    with callbacks.collect_runs() as cb:
```
jordanrfrazier, line 245: Split these up because it's easier. You could probably do something like `len` of runs after first invocation, then get `traced_runs[len]` for second invocation, but this worked so left it as is.

PR #250, opened 2024-01-29
[ref] simplify astra collection cleanup

jordanrfrazier:
We still see some errors like
```
ValueError: API Exception while running bulk insertion: {'errors': [{'message': 'Collection does not exist, collection name: documents_b00f7c83', 'errorCode': 'COLLECTION_NOT_EXIST'}]}

```

I don't think this PR will fix those (if they even are a result of our test cleanup logic), but this should simplify the cleanup a bit anyway, and might help since it removes at least one unnecessary delete_collection call.

### DIFF
--- a/ragstack-e2e-tests/e2e_tests/test_utils/astradb_vector_store_handler.py
+++ b/ragstack-e2e-tests/e2e_tests/test_utils/astradb_vector_store_handler.py
@@ -201,23 +201,19 @@ class AstraDBVectorStoreHandler(VectorStoreHandler):
             self.try_delete_with_backoff(collection, sleep * 2, max_tries)
 
     def ensure_astra_env_clean(self, blocking=False):
-        logging.info("Ensuring astra env is clean")
-        self.delete_collection_handler.run_delete(self.astra_ref.collection)
         collections = (
             self.default_astra_client.get_collections().get("status").get("collections")
         )
-        logging.info(f"Existing collections: {collections}")
+        logging.info("Deleting all existing collections: {collections}")
         for name in collections:
-            if name == self.astra_ref.collection:
-                continue
             self.delete_collection_handler.run_delete(name)
+
         if blocking:
             self.delete_collection_handler.await_ongoing_deletions_completed()
 
     def before_test(self) -> AstraDBVectorStoreTestContext:
         super().check_implementation()
         self.ensure_astra_env_clean(blocking=True)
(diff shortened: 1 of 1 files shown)


ragstack-e2e-tests/e2e_tests/test_utils/astradb_vector_store_handler.py
```diff
@@ -201,23 +201,19 @@ def try_delete_with_backoff(self, collection: str, sleep=1, max_tries=5):
             self.try_delete_with_backoff(collection, sleep * 2, max_tries)
 
     def ensure_astra_env_clean(self, blocking=False):
-        logging.info("Ensuring astra env is clean")
-        self.delete_collection_handler.run_delete(self.astra_ref.collection)
```
jordanrfrazier, line 205: Actually, was this (incorrectly) deleting the collection we just created for the purposes of the test?

And later on when we reassigned the `self.astra_ref.collection = .. `, how is that new collection being created?

ragstack-e2e-tests/e2e_tests/test_utils/astradb_vector_store_handler.py
```diff
@@ -233,4 +229,4 @@ def before_test(self) -> AstraDBVectorStoreTestContext:
         return AstraDBVectorStoreTestContext(self)
 
     def after_test(self):
-        self.ensure_astra_env_clean(blocking=False)
+        self.delete_collection_handler.run_delete(self.astra_ref.collection)
```
nicoloboschi, line 232: some tests use more than one collection, for example for the chat memory

ragstack-e2e-tests/e2e_tests/test_utils/astradb_vector_store_handler.py
```diff
@@ -201,23 +201,19 @@ def try_delete_with_backoff(self, collection: str, sleep=1, max_tries=5):
             self.try_delete_with_backoff(collection, sleep * 2, max_tries)
 
     def ensure_astra_env_clean(self, blocking=False):
-        logging.info("Ensuring astra env is clean")
```
nicoloboschi, line 204: the problem with this approach is that if a collection deletion is in progress, the get_collections won't return it. so we could end up with the same issue we have now

nicoloboschi (changes requested):

PR #263, opened 2024-02-01
[RAG-176] - Docs - Add hotels app

mendonk:
* The GitPod and repo are linked to @hemidactylus repo. Not sure if that's OK or not, but it does work.
* App runs fine with RAGStack 0.6.0 instead of langchain, but I had to downgrade pydantic to 1.10.14 to use the LLM in the app (ie no hotel summaries in the page). It looked like pydantic was confused about the community package imports.


docs/modules/examples/pages/hotels-app.adoc
```diff
@@ -0,0 +1,133 @@
...
+[source,curl]
+----
+curl -v -s --location \
+--request POST "https://ASTRA_DB_ID-ASTRA_DB_REGION.apps.astra.datastax.com/api/json/v1/default_keyspace" \
+--header "X-Cassandra-Token: AstraCS:..." \
+--header "Content-Type: application/json" \
+--header "Accept: application/json" \
+--data '{
+  "deleteCollection": {
+    "name": "hotels"
```
hemidactylus, line 130: Actually the app creates five collections (`hotels`, `cities`, `reviews`, `hotel_reviews_embeddings`, `users`).
A proper cleanup would have the user drop them all.

Maybe better to wrap this in a clealy labeled tiny python script "setup/cleanup_delete_everything.py" or something. (this would be on me in the repo).

Not to mention the _sixth_ collection (LLM cache) that will also the party as soon as the 5-limit is lifted...

hemidactylus:
LGTM as is, but I would try to spend some time improving the demo repo (at which point it would be good to come back here to the docs):
- transferring to legit GH org
- modernize date support

Also, even more compelling, I could make the repo a RAGStack-first demo, why not? I mean, a future version would have (a pinned version of) RAGStack in its requirements.txt, no need to give instructions along the lines of "hey, ignore requirements.txt and do this". How does that sound? Also this would open the way to a RAGStack-flavored Open-in-Gitpod zero-hassle experience, a big win IMO.

hemidactylus (approved):

docs/modules/examples/pages/hotels-app.adoc
mendonk: Good call. 
Here's a PR to add a delete_collection script:
https://github.com/hemidactylus/langchain-astrapy-hotels-app/pull/1

mendonk:
> * transferring to legit GH org

I can take this one if you want- just a transfer over here to https://github.com/orgs/datastaxdevs/repositories

> * modernize date support
> * ragstack first demo 

These would be awesome, thanks! If you have questions about cutting over to ragstack just let me know.
In the meantime, I'm going to publish this - it works and we can iterate.

PR #267, opened 2024-02-02
[RAG-178] - Docs - MMR 

mendonk:
* Adapt [this example ](https://cassio.org/frameworks/langchain/qa-maximal-marginal-relevance/) to ragstack and document it. 
* Works
`(mmr-test) ➜  mmr-notebook-test python3 mmr-test-final.py
92113d9999ad463a8ad6aef0b2181f70
840d924b3870412cbf1101a0f6da0e02
0a4977930c134df2a11d798c42470792
57cd5e3560824500a4b7f2e242191742
2d9d171d704c48c89cc81e9812008391
Similarity-based chain:
  ANSWER : The party was thrown by all the toads and frogs under the moon that night.
  SOURCES: Barney's story at the pub
MMR-based chain:
  ANSWER : The party that night was thrown by the frogs and toads, and the rule was for everyone to wear a purple hat.
  SOURCES: Barney's story at the pub, The chronicles at the village library`

### DIFF
--- a/examples/notebooks/qa-maximal-marginal-relevance.ipynb
+++ b/examples/notebooks/qa-maximal-marginal-relevance.ipynb
@@ -1,0 +1,690 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "005e19ad",
+   "metadata": {},
+   "source": [
+    "# VectorStore/QA, MMR support"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "0778d5f6",
+   "metadata": {},
+   "source": [
+    "_**NOTE:** this uses Cassandra's \"Vector Search\" capability.\n",
+    "Make sure you are connecting to a vector-enabled database for this demo._"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "102fec23",
+   "metadata": {},
+   "source": [
+    "Cassandra's `VectorStore` allows for Vector Search with the **Maximal Marginal Relevance (MMR)** algorithm.\n",
+    "\n",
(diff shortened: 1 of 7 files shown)


examples/notebooks/qa-maximal-marginal-relevance.ipynb
```diff
@@ -0,0 +1,690 @@
...
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "LLM+embeddings from OpenAI\n"
+     ]
+    }
+   ],
+   "source": [
+    "import os\n",
+    "from llm_choice import suggestLLMProvider\n",
```
hemidactylus, line 148: This is machinery that in the original was kept in a separate module (which I don't see here?)
Also it is quite daunting (I sort of regret having given too much weight to "let's support various LLMs/embeddings"). I see it's been simplified to a vanilla OpenAI LLM instantiation (one line!) in the adoc.

1. could the same simplification occur here, perhaps?
2. I suspect this would not work as is since there's no "llm_choice.py". One may be better off with the "colab-ready" version, which through black magic is derived from this one in a way that can run self-containedly: https://github.com/CassioML/cassio-website/blob/main/docs/frameworks/langchain/.colab/colab_qa-maximal-marginal-relevance.ipynb

hemidactylus: Come to think of it, the "open in colab" link at the top of [this page](https://cassio.org/frameworks/langchain/qa-maximal-marginal-relevance/) is good as is - except probably you'll want to replace the langchain deps with ragstack

mendonk: 1. I will apply these changes to this notebook in another PR. 
2. I added the llm_choice and cqlsession dependencies, and it passes CI so I'm merging. (Notebook doesn't have nbmake points to fail CI). I can revert to the other link if it doesn't work.

PR #272, opened 2024-02-05
[tests] Moves lint to separate job (and formats yml)

jordanrfrazier:
Moved `lint` to a prereq job so that e2e will not run if lint fails. 

Also uses `redhat.vscode-yaml` from the vscode RedHat`YAML` extension for formatting of yml (https://github.com/redhat-developer/vscode-yaml/blob/e5a67a112a9a52eb8d8c991bf444fb309b2eb913/README.md) - @nicoloboschi If you have another style preference, I can swap to that.


jordanrfrazier:
Closing -- going to go with `yamllint` for the non-vscode users.

PR #275, opened 2024-02-06
[ref] use singleton for vector store handler

jordanrfrazier:
MyPy work was complaining about this are a bit, so I went on a tangent. Cleans up the singleton handler a bit so it doesn't have to use shared class variables and only calls `init` once. 

Though, this PR doesn't do much actual cleanup aside from that. I think it's a good step and later on can clean up the `VectorStoreImplementation` stuff.

### DIFF
--- a/ragstack-e2e-tests/e2e_tests/test_utils/astradb_vector_store_handler.py
+++ b/ragstack-e2e-tests/e2e_tests/test_utils/astradb_vector_store_handler.py
@@ -25,6 +25,7 @@ from e2e_tests.test_utils.cassandra_vector_store_handler import (
     EnhancedCassandraLangChainVectorStore,
     CassandraChatMessageHistory,
 )
+from e2e_tests.test_utils.singleton import Singleton
 
 from astrapy.db import AstraDB as AstraPyClient
 
@@ -169,13 +170,18 @@ class AstraDBVectorStoreTestContext(VectorStoreTestContext):
                 table_name=self.handler.collection_name,
                 **kwargs,
             )
-        else:
+        elif self.handler.implementation == VectorStoreImplementation.ASTRADB:
             vector_store = EnhancedAstraDBLangChainVectorStore(
                 collection_name=self.handler.collection_name,
                 token=self.handler.token,
                 api_endpoint=self.handler.api_endpoint,
                 **kwargs,
             )
+        else:
+            raise ValueError(
+                "Invalid vstore implementation: {self.handler.implementation}"
+            )
+
         logging.info("Created vector store")
         return vector_store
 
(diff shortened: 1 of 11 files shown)


ragstack-e2e-tests/e2e_tests/test_utils/vector_store_handler.py
```diff
@@ -51,18 +52,22 @@ def new_llamaindex_vector_store(self, **kwargs) -> EnhancedLlamaIndexVectorStore
 class VectorStoreHandler(ABC):
     def __init__(
         self,
-        implementation: VectorStoreImplementation,
         supported_implementations: List[VectorStoreImplementation],
     ):
-        self.implementation = implementation
+        self.implementation = VectorStoreImplementation.UNINITIALIZED
```
jordanrfrazier, line 57: I'd say this is the downside with this refactoring, but I think it suggests that we should have a separate implementation for the astra client and cassio

ragstack-e2e-tests/e2e_tests/test_utils/astradb_vector_store_handler.py
```diff
@@ -169,13 +170,18 @@ def new_langchain_vector_store(self, **kwargs) -> EnhancedLangChainVectorStore:
...
+        elif self.handler.implementation == VectorStoreImplementation.ASTRADB:
             vector_store = EnhancedAstraDBLangChainVectorStore(
                 collection_name=self.handler.collection_name,
                 token=self.handler.token,
                 api_endpoint=self.handler.api_endpoint,
                 **kwargs,
             )
+        else:
+            raise ValueError(
+                "Invalid vstore implementation: {self.handler.implementation}"
```
nicoloboschi, line 182:
```suggestion
                f"Invalid vstore implementation: {self.handler.implementation}"
```

ragstack-e2e-tests/e2e_tests/test_utils/vector_store_handler.py
```diff
@@ -51,18 +52,22 @@ def new_llamaindex_vector_store(self, **kwargs) -> EnhancedLlamaIndexVectorStore
...
         self.supported_implementations = supported_implementations
 
     def check_implementation(self):
+        if self.implementation == VectorStoreImplementation.UNINITIALIZED:
```
nicoloboschi, line 61: is this ever called ?

ragstack-e2e-tests/e2e_tests/test_utils/vector_store_handler.py
nicoloboschi: not really, the fact is that AstraDB works with using astrapy both in cassandra and astradb mode

jordanrfrazier: Yes, but it remains that cassandra mode requires some additional setup. If the shared functionality remains in the `AstraDBVectorStoreHandler`, an `CassIOAstraDBVectorStoreHandler` (naming tbd) can handle that additional setup. 

That would let us do 
```
  if vector_database_type == "astradb":
        return AstraDBVectorStoreHandler()
   elif vector_database_type == "astradb-cassio":
      return CassIOAstraDBVectorStoreHandler()
    elif vector_database_type == "local-cassandra":
        return CassandraVectorStoreHandler()
```

and then test fixtures will not need to specify `Implementation`. 

```
@pytest.fixture
def astra_db():
    handler = get_vector_store_handler()
    context = handler.before_test()
    yield context
    handler.after_test()
```

It then splits up the e2e tests in CI between cassandra and data API modes, which I see as beneficial. 

But, if you disagree, I can work on fixing the singleton in another manner. This was a blocker on the `MyPy` work from last week, as it wanted some refactoring in this area in particular.

PR #282, opened 2024-02-13
[RAG-201] - LlamaIndex migration

mendonk:
* Add llama migration instructions - (these will probably change based on the ragstack package update)
* Move changelog in nav for clarity @nicoloboschi 
* Rename llama-parse "Export" step that actually uses dotenv

### DIFF
--- a/docs/modules/ROOT/pages/migration.adoc
+++ b/docs/modules/ROOT/pages/migration.adoc
@@ -1,35 +1,34 @@
 = Migrate to RAGStack
 
-Migrating an existing LangChain application to RAGStack is easy - just change your `requirements.txt` or `pyproject.toml` file to use `ragstack-ai`.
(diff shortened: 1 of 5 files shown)


docs/modules/ROOT/pages/migration.adoc
```diff
@@ -211,7 +210,200 @@ yarl                1.9.4
...
+====
+
+. This application requires installation of the following packages:
++
+[source,python]
+----
+pip install llama-index
+----
++
+. Your application is tested and working at `llama-index` version `0.9.29`. But then, LlamaIndex version `0.10.1` changes the module to split every integration into its own PyPi package. Oh no, your application no longer works!
```
nicoloboschi, line 286: to be fair here, users would have stick their dependency to 0.9.x so they wouldn't be affected by this change implicitly.

Also ragstack now can select specific integrations to be included in the stack but I believe we won't add anything by default except for astra (and maybe openai)

nicoloboschi (approved):
@mendonk I've left one comment but we can open another PR later, it's not necessary at the moment, I'll merge this one

PR #327, opened 2024-03-13
base colbert implementation

zzzming:
A ColBERT base implementation that has no LangChain and LlamaIndex implementation. This module only does
* ColBERT based embedding
* Create Astra DB vector tables
* Load embedding to AstraDB tables
* Retrieve from AstraDB and compute based on ColBERT recall methods
* including an integration test

We will add LangChain integration in another module.


ragstack/colbertbase/colbertbase/astra_db.py
```diff
@@ -0,0 +1,179 @@
...
+
+        self.session.execute("""
+            CREATE TABLE IF NOT EXISTS colbert_embeddings (
+                title text,
+                part int,
+                embedding_id int,
+                bert_embedding vector<float, 128>,
+                PRIMARY KEY (title, part, embedding_id)
+            );
+        """)
```
michaelsembwever, line 117: it is possible to collapse these into one table
see https://github.com/datastax-labs/colbert-wikipedia-data/blob/main/schema.cql

ragstack/colbertbase/colbertbase/astra_retriever.py
```diff
@@ -0,0 +1,140 @@
...
+            embeddings_for_part = [tensor(row.bert_embedding) for row in rows]
+            # score based on The function returns the highest similarity score
+            #(i.e., the maximum dot product value) between the query vector and any of the embedding vectors in the list.
+            scores[(title, part)] = sum(max_similarity_torch(qv, embeddings_for_part, self.is_cuda) for qv in query_encodings)
+        # load the source chunk for the top k documents
+        docs_by_score = sorted(scores, key=scores.get, reverse=True)[:k]
+        answers = []
+        rank = 1
+        for title, part in docs_by_score:
+            rs = self.astra.session.execute(self.astra.query_part_by_pk_stmt, [title, part])
```
michaelsembwever, line 134: i think the retriever approach i took here is faster:
https://github.com/datastax-labs/colbert-wikipedia-data/blob/main/serve.py#L52-L80

michaelsembwever:
if this is cql based shouldn't it be "cassandra" ? 
and for astra we use the data api ?

ragstack/colbertbase/colbertbase/astra_retriever.py
```diff
@@ -0,0 +1,140 @@
...
+        # the min of query_maxlen is 32
+        top_k = max(math.floor(len(query_encodings) / 2), 16)
+        if self.verbose:
+            print(f"Total number of chunks: {count}, query length {len(query)} embeddings top_k: {top_k}")
+
+        # find the most relevant documents
+        docparts = set()
+        for qv in query_encodings:
+            # per token based retrieval
+            rows = self.astra.session.execute(self.astra.query_colbert_ann_stmt, [list(qv), top_k])
```
michaelsembwever, line 118: perf gain to be had here with concurrent async queries…

ragstack/colbertbase/colbertbase/astra_db.py
zzzming: @michaelsembwever in your `articles` table, is the `bert_embedding` per token's embedding? If so, would body per token or per chunk?

michaelsembwever: > is the bert_embedding per token's embedding? 

yes

> If so, would body per token or per chunk?

body per chunk.
there's still the separate query to fetch the body.  the body is found in the rows where `bert_embedding_no=-1`

note, i probably could have saved a bit by using smallint for bert_embedding_no instead of int.  same goes for chunk_no.  we don't expect more than 32k for each.

zzzming: So the body is duplicated for each token. That's a lot of duplicated data since even with the default chunk size. There could be 160 to 220 tokens per chunk.

ragstack/colbertbase/colbertbase/astra_retriever.py
zzzming: That's a great comments. execute_async improved about 70% on 50000 chunks search.

zzzming: Good point! I changed to use futures and execute_async. Since Eq is much less than Ed, the improvement is still much better around 20% latency improvement

ragstack/colbertbase/colbertbase/astra_retriever.py
zzzming: Good point! I changed to use futures and execute_async. Since Eq is much less than Ed, the improvement is still much better around 20% latency improvement

ragstack/colbertbase/pyproject.toml
```diff
@@ -0,0 +1,22 @@
+[tool.poetry]
+name = "colbertbase"
```
cbornet, line 2: We said we would not create a separate package atm ?

.github/workflows/modules.yml
```diff
@@ -0,0 +1,57 @@
...
+        run: |
+          curl -sSL https://install.python-poetry.org | python3 -
+          echo "$HOME/.local/bin" >> $GITHUB_PATH
+
+      - name: Install dependencies
+        run: |
+          cd ragstack/colbertbase
+          poetry install
+
+      - name: Lint with flake8
```
cbornet, line 49: Can we use `black` instead

ragstack/colbertbase/colbertbase/token_embedding.py
```diff
@@ -0,0 +1,114 @@
...
+    
+    def get_all_token_embeddings(self) -> List[PerTokenEmbeddings]:
+        return self.__token_embeddings
+
+    def get_text(self):
+        return self.__text
+    
+#
+# This is the base class for token based embedding
+# ColBERT token embeddings is an example of a class that inherits from this class
```
cbornet, line 104: If there's comment to do on the class, better put it in the docstring

ragstack/colbertbase/colbertbase/astra_db.py
michaelsembwever: No, there's no duplication. 

Example data: 

```
EXPAND set to ON
cqlsh:wikidata> select * from articles limit 3;

----------------------------+
@ Row 1
----------------------------+
wiki                       | simplewiki
 language                   | en
 title                      | Luciano Spinosi
 chunk_no                   | -1
 bert_embedding_no          | -1
 all_minilm_l6_v2_embedding | null
 bert_embedding             | null
 body                       | Luciano Spinosi\n\nLuciano Spinosi (born 9 May, 1950) is a former Italian football player. He has played for Italy national team. Strack-Zimmermann, Benjamin. "Luciano Spinosi". www.national-football-teams.com. "Luciano Spinosi - International Appearances". www.rsssf.com. This short article about a\xa0sports person can be made longer. You can help Wikipedia by adding to it.
 id                         | 182678
 revision                   | 8979208

----------------------------+
@ Row 2
----------------------------+
wiki                       | simplewiki
 language                   | en
 title                      | Luciano Spinosi
 chunk_no                   | 0
 bert_embedding_no          | -1
 all_minilm_l6_v2_embedding | [<vanilla_embedding_for_chunk_0]
 bert_embedding             | null
 body                       | Luciano Spinosi\n\nLuciano Spinosi (born 9 May, 1950) is a former Italian football player. He has played for Italy national team. Strack-Zimmermann, Benjamin. "Luciano Spinosi". www.national-football-teams.com. "Luciano Spinosi - International Appearances". www.rsssf.com. This short article about a\xa0sports person can be made longer. You can help Wikipedia by adding to it.
 id                         | 182678
 revision                   | 8979208

----------------------------+
@ Row 3
----------------------------+
 wiki                       | simplewiki
 language                   | en
 title                      | Luciano Spinosi
 chunk_no                   | 0
 bert_embedding_no          | 0
 all_minilm_l6_v2_embedding | null
 bert_embedding             | [<colbert_embedding_for_chunk_0_token_0>]
 body                       | null
 id                         | 182678
 revision                   | 8979208
```

.github/workflows/modules.yml
cbornet: We should also use ruff and mypy with strict type checking.

ragstack/colbertbase/colbertbase/astra_db.py
```diff
@@ -0,0 +1,177 @@
...
+
+        if keyspace not in self.cluster.metadata.keyspaces.keys():
+            raise ValueError(f"Keyspace '{keyspace}' does not exist. please create it first.")
+
+        self.create_tables()
+
+        # prepare statements
+
+        chunk_counts_cql = f"""
+        SELECT COUNT(*) FROM {keyspace}.chunks
```
michaelsembwever, line 51: this won't work in any real system

ragstack/colbertbase/colbertbase/astra_colbert_embedding.py
```diff
@@ -0,0 +1,224 @@
...
+            nranks: int = -1,
+            query_maxlen: int = 32,
+            **data: Any,
+    ):
+        self.__cuda = torch.cuda.is_available()
+        total_visible_gpus = 0
+        if self.__cuda:
+            self.__cuda_device_count = torch.cuda.device_count()
+            self.__cuda_device_name = torch.cuda.get_device_name()
+            print(f"nrank {nranks}")
```
cbornet, line 86: Use logging instead

ragstack/colbertbase/colbertbase/astra_colbert_embedding.py
```diff
@@ -0,0 +1,224 @@
...
+            self,
+            query: Union[str, List[str]],
+            full_length_search: bool = False,
+            query_maxlen: int = 32,
+        ):
+        queries = query if type(query) is list else [query]
+        bsize = 128 if len(queries) > 128 else None
+
+        self.checkpoint.query_tokenizer.query_maxlen = max(query_maxlen, self.colbert_config.query_maxlen)
+        Q = self.checkpoint.queryFromText(queries, bsize=bsize, to_cpu=True, full_length_search=full_length_search)
```
cbornet, line 140: change var name to `query`

ragstack/colbertbase/colbertbase/astra_db.py
```diff
@@ -0,0 +1,177 @@
...
+        delete_existed_passage: bool = False
+    ) -> None:
+        if delete_existed_passage:
+            for p in embeddings:
+                try:
+                    self.delete_title(p.title())
+                except Exception as e:
+                    # no need to throw error if the title does not exist
+                    # let the error propagate
+                    print(f"delete title {p.title()} error {e}")
```
nicoloboschi, line 154: > no need to throw error if the title does not exist

if the title doens't exists, you shouldn't have any error from the DELETE statement. We should instead raise error if something is not working in the session

ragstack/colbertbase/colbertbase/astra_db.py
```diff
@@ -0,0 +1,177 @@
...
+        # insert chunks
+        p_parameters = [(p.title(), p.part(), p.get_text()) for p in embeddings]
+        execute_concurrent_with_args(self.session, self.insert_chunk_stmt, p_parameters)
+        if (self.verbose):
```
nicoloboschi, line 158: I'd remove the verbose parameter and play with logging levels

ragstack/colbertbase/colbertbase/astra_db.py
```diff
@@ -0,0 +1,177 @@
...
+            for p in embeddings:
+                try:
+                    self.delete_title(p.title())
+                except Exception as e:
+                    # no need to throw error if the title does not exist
+                    # let the error propagate
+                    print(f"delete title {p.title()} error {e}")
+        # insert chunks
+        p_parameters = [(p.title(), p.part(), p.get_text()) for p in embeddings]
+        execute_concurrent_with_args(self.session, self.insert_chunk_stmt, p_parameters)
```
nicoloboschi, line 157: is this synchronous ?

ragstack/colbertbase/colbertbase/astra_retriever.py
nicoloboschi: +1 for using async

ragstack/colbertbase/colbertbase/token_embedding.py
```diff
@@ -0,0 +1,114 @@
...
+    __title: str
+    __id: uuid.UUID
+
+    def __init__(
+            self,
+            text: str,
+            title: str = "",
+            part: int = 0,
+            id: uuid.UUID = None,
+            model: str = "colbert-ir/colbertv2.0",
```
nicoloboschi, line 52: DEFAULT_COLBERT_MODEL ?

ragstack/colbertbase/colbertbase/token_embedding.py
```diff
@@ -0,0 +1,114 @@
...
+            dim: int = 128,
```
nicoloboschi, line 53: DEFAULT_COLBERT_DIM?

ragstack/colbertbase/pyproject.toml
nicoloboschi: yes exactly, we don't need another pyyproject for now, let's add the dependencies in the main ragstack pyproject. Note that CassIO is already available so there's no need add `cassandra-driver`

ragstack/colbertbase/tests/__init__.py
```diff
@@ -0,0 +1 @@
+# place holder
```
nicoloboschi, line 1: remove

nicoloboschi (changes requested):
Nice work @zzzming 

Some general comments: 
- use logging library for logs
- move everything under the ragstack-ai module
- I'm fine in committing this without CassIO but we need to switch to it before cutting a release
- We should have integration tests with astra/dse (in another PR). I can help you with that later 

Additionaly, there are some libraries that are required and others are optional:
required:
- colbert
- pyarrow
- torch 
Optional 
- psutil
- faissgpu

We should think about the size of these and maybe put them in an extra

ragstack/colbertbase/colbertbase/astra_db.py
```diff
@@ -0,0 +1,177 @@
...
+            );
+        """)
+        print("Created chunks table")
+
+        self.session.execute("""
+            CREATE TABLE IF NOT EXISTS colbert_embeddings (
+                title text,
+                part int,
+                embedding_id int,
+                bert_embedding vector<float, 128>,
```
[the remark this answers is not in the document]
michaelsembwever: this goes to my question about `astra_db.py` vs `cassandra-db.py`  ? 

this PR is implementing the cql approach.
the data api approach would be appropriate for `astra_db.py`

ragstack/colbertbase/colbertbase/astra_db.py
```diff
@@ -0,0 +1,177 @@
...
+def required_cred(cred: str):
+    if cred is None or cred == "":
+        raise ValueError("Please provide credentials")
+
+class AstraDB:
+    def __init__(
+            self,
+            secure_connect_bundle: str="",
+            astra_token: str=None,
+            keyspace: str="colbert128",
```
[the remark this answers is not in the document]
michaelsembwever: see https://github.com/datastax/ragstack-ai/pull/327#discussion_r1524527282

ragstack/colbertbase/colbertbase/astra_db.py
```diff
@@ -0,0 +1,177 @@
...
+from cassandra.concurrent import execute_concurrent_with_args
+
+from .token_embedding import PassageEmbeddings
+
+def required_cred(cred: str):
+    if cred is None or cred == "":
+        raise ValueError("Please provide credentials")
+
+class AstraDB:
+    def __init__(
```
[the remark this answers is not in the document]
michaelsembwever: what's the advantage of cassio ? 
(and if using cassandra-driver do we still need secure bundle for astra vector compatibility?)

cbornet:
Are we good with pulling the deps required by colbert by default ? Or should we put them in an extra ?

ragstack/colbertbase/colbertbase/astra_db.py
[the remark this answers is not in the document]
nicoloboschi: we want a single auth interface. Since it's likely that you will use colbert with the langchain/llama CassandraStore, you will likely call `cassio.init`. If we can leverage it here, it'd be better

ragstack/colbertbase/colbertbase/astra_db.py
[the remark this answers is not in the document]
nicoloboschi: right but actually if we only use CQL (for now) we have no clue if the target is an Astra database or not. 
I think we should have only one solution that works with Cassandra, I'm fine with bert_embedding but $vector makes more sense to me in general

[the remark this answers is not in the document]
michaelsembwever: yes.

if it's cql then it should be cql compatible with all Astra/DSE/Cassandra,  named `cassandra-db.py` and not use secure bundle.

`astra-db.py` should then be the DATA API variant , which is only compatible with Astra (and soon DSE).

my understanding is that from a product perspective the DATA API variant takes higher priority (in lei of any customers prioritising the cql variant).

ragstack/colbertbase/colbertbase/astra_retriever.py
cbornet: And to add to this, we'll need an asyncio async/await impl at some point.

ragstack/colbertbase/colbertbase/astra_db.py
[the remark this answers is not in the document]
cbornet: That's were cassio will help as it already supports the diffs between Cassandra and Astra DB

ragstack/colbertbase/colbertbase/astra_db.py
zzzming: Can you explain why it's won't in any real system? Is it because too many chunks?

ragstack/colbertbase/colbertbase/astra_colbert_embedding.py
zzzming: Q and D is a standard ColBERT term. If someone read the ColBERT paper, Q is the term being used.

ragstack/colbertbase/colbertbase/astra_colbert_embedding.py
```diff
@@ -0,0 +1,224 @@
...
+        print("creating checkpoint")
+        self.checkpoint = Checkpoint(self.colbert_config.checkpoint, colbert_config=self.colbert_config)
+        self.encoder = CollectionEncoder(config=self.colbert_config, checkpoint=self.checkpoint)
+        self.__cuda = torch.cuda.is_available()
+        if self.__cuda:
+            self.checkpoint = self.checkpoint.cuda()
+
+        self.print_memory_stats("ColbertTokenEmbeddings")
+
+    def embed_documents(self, texts: List[str], title: str="") -> List[PassageEmbeddings]:
```
epinzur, line 121: would you consider changing `title` to `id`?

ragstack/colbertbase/colbertbase/astra_colbert_embedding.py
```diff
@@ -0,0 +1,224 @@
...
+        if title == "":
+            title = str(uuid.uuid4())
+        """Embed search docs."""
+        return self.encode(texts, title)
+
+    def embed_query(self, text: str, title: str) -> PassageEmbeddings:
```
epinzur, line 127: i'm confused why a query would have a `title` param.

ragstack/colbertbase/colbertbase/token_embedding.py
```diff
@@ -0,0 +1,114 @@
...
+    __text: str
+    __title: str
+    __id: uuid.UUID
+
+    def __init__(
+            self,
+            text: str,
+            title: str = "",
+            part: int = 0,
+            id: uuid.UUID = None,
```
epinzur, line 51: what is difference between `title` and `id`?

ragstack/colbertbase/colbertbase/astra_colbert_embedding.py
```diff
@@ -0,0 +1,224 @@
...
+
+        if condition:
+            msg = msg if not pad else f'\n{msg}\n'
+            print(msg, flush=True)
+
+        return msg
+
+    def print_memory_stats(self, message=''):
+        try:
+            import psutil  # Remove before releases? Or at least make optional with try/except.
```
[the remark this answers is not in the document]
epinzur: seems stats like this would probably be generated by some external tool, and not be in ragstack

ragstack/colbertbase/colbertbase/astra_colbert_embedding.py
```diff
@@ -0,0 +1,224 @@
+from typing import Any, Dict, List, Union
```
epinzur, line 1: seems like this file isn't astra related. should we rename to `colbert_token_embeddings.py`?

epinzur: or just `colbert_embedding.py`

zzzming:
> if this is cql based shouldn't it be "cassandra" ? and for astra we use the data api ?

renamed

ragstack/colbertbase/pyproject.toml
zzzming: refactored

.github/workflows/modules.yml
zzzming: It uses the existing black and ruff.

ragstack/colbertbase/colbertbase/token_embedding.py
zzzming: fixed

ragstack/colbertbase/colbertbase/astra_colbert_embedding.py
```diff
@@ -0,0 +1,224 @@
...
+    __accumsteps: int = 1,
+    __lr: float = 0.000003,
+    __maxsteps: int = 500000,
+    __nway: int = 2,
+    __use_ib_negatives: bool = False,
+    __reranker: bool = False,
+    __is_cuda: bool = False
+
+    @classmethod
+    def validate_environment(self, values: Dict) -> Dict:
```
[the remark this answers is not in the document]
zzzming: removed

ragstack/colbertbase/colbertbase/astra_db.py
zzzming: Removed. I don't need this. Because it's only needed to reduce k when k is less than the total row. K is 10, any meaningful even test data should have more than 10 chunks.

ragstack/colbertbase/colbertbase/astra_colbert_embedding.py
zzzming: fixed

.github/workflows/modules.yml
```diff
@@ -0,0 +1,57 @@
+#
```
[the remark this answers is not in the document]
zzzming: Removed after code refactoring

ragstack/colbertbase/.flake8
```diff
@@ -0,0 +1,3 @@
+[flake8]
```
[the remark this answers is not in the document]
zzzming: done using the default black and ruff

ragstack/colbertbase/README.md
```diff
@@ -0,0 +1,14 @@
+# ColBERT retrieval
```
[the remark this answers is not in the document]
zzzming: done

ragstack/colbertbase/README.md
```diff
@@ -0,0 +1,14 @@
+# ColBERT retrieval 
+
+This is a ColBERT retrieval based on Astra DB or Cassandra implementation.
+
+This module only depends on ColBERT and Cassandra driver with no LangChain and LlamaIndex dependencies.
+
+## Examples
```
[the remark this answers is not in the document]
zzzming: removed

ragstack/colbertbase/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+
+## Env
+```bash
+poetry install
```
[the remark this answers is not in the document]
zzzming: removed

ragstack/colbertbase/colbertbase/astra_colbert_embedding.py
zzzming: This is not done. I will address later

ragstack/colbertbase/colbertbase/astra_colbert_embedding.py
zzzming: Removed. Query does not need title. 
Actually, this piece of code was not used for query. But I made a unit test for this. Since this query embedding does not have padding, I have not tested it.

ragstack/colbertbase/colbertbase/token_embedding.py
zzzming: Title should be supplied by the user or specified as metadata.
Id is to track the parent of each per token embeddings which is the chunk.
I don't use Id at the moment. So it can be deleted. I have not removed them yet.

ragstack/colbertbase/colbertbase/astra_colbert_embedding.py
zzzming: Renamed to colbert_embedding.py

zzzming:
> Are we good with pulling the deps required by colbert by default ? Or should we put them in an extra ?

I put them in extra [colbert]

ragstack/colbertbase/colbertbase/astra_db.py
zzzming: If a single table is requirement, there might be a way to get the actual text for the token from the ColBERT library. I have seen some functions in library that might be able to return the token. 

I am just not sure just the same table to store two different information makes sense. Is this a good practice? 
May I suggest I take this as a separate activity?

ragstack/colbert/colbert_embedding.py
```diff
@@ -0,0 +1,178 @@
...
+        )
+        self.encoder = CollectionEncoder(
+            config=self.colbert_config, checkpoint=self.checkpoint
+        )
+        self.__cuda = torch.cuda.is_available()
+        if self.__cuda:
+            self.checkpoint = self.checkpoint.cuda()
+
+    def embed_documents(
+        self, texts: List[str], title: str = ""
```
epinzur, line 110: can you change `title` to `id` throughout this file?

ragstack/colbert/astra_retriever.py
```diff
@@ -0,0 +1,162 @@
...
+        for qv in query_encodings:
+            # per token based retrieval
+            doc_future = self.astra.session.execute_async(
+                self.astra.query_colbert_ann_stmt, [list(qv), top_k]
+            )
+            doc_futures.append(doc_future)
+
+        for future in doc_futures:
+            rows = future.result()
+            docparts.update((row.title, row.part) for row in rows)
```
epinzur, line 120: change `title` to `id` throughout?

ragstack/colbert/token_embedding.py
```diff
@@ -0,0 +1,120 @@
...
+
+class PerTokenEmbeddings:
+    __embeddings: List[float]
+
+    def __init__(
+        self,
+        id: int,
+        part: int,
+        parent_id: uuid.UUID = None,
+        title: str = "",
```
epinzur, line 18: can we drop `title` here?

ragstack/colbert/token_embedding.py
```diff
@@ -0,0 +1,120 @@
...
+        return self.parent_id
+
+    def part(self):
+        return self.part
+
+
+class PassageEmbeddings:
+    __token_embeddings: List[PerTokenEmbeddings]
+    __text: str
+    __title: str
```
epinzur, line 45: can we drop `__title` from this class?

tests/unit-tests/test_astra_embedding_retrieval.py
```diff
@@ -0,0 +1,112 @@
...
+        return chunks
+
+    # Generate the chunks based on the narrative
+    chunks = generate_chunks(narrative, chunk_size, overlap_size)
+
+    # Output the first few chunks to ensure they meet the specifications
+    for i, chunk in enumerate(chunks[:3]):  # Displaying the first 3 chunks for brevity
+        print(f"Chunk {i+1}:\n{chunk}\n{'-'*50}\n")
+
+    title = "Marine Animals habitat"
```
epinzur, line 50: move saving title into the document metadata

tests/unit-tests/test_colbert_embeddings.py
```diff
@@ -0,0 +1,66 @@
...
+    passagesEmbeddings = colbert.embed_documents(["test1", "test2"])
+
+    assert len(passagesEmbeddings) == 2
+
+    assert passagesEmbeddings[0].get_text() == "test1"
+    assert passagesEmbeddings[1].get_text() == "test2"
+
+    # generate uuid based title
+    assert passagesEmbeddings[0].title() != ""
+    assert passagesEmbeddings[1].title() != ""
```
epinzur, line 20: change to testing `id`

pyproject.toml
```diff
@@ -24,10 +24,14 @@ langchain-openai = "0.0.8"
 langchain-google-genai = { version = "0.0.9", optional = true }
 langchain-google-vertexai = { version = "0.1.0", optional = true }
 langchain-nvidia-ai-endpoints = { version = "0.0.3", optional = true }
+colbert-ai = { version = "0.2.19", optional = true }
+pyarrow = "14.0.1"
```
nicoloboschi, line 28: optional ?

pyproject.toml
```diff
@@ -24,10 +24,14 @@ langchain-openai = "0.0.8"
...
+torch = { version = "2.2.1", optional = true }
 
 [tool.poetry.extras]
 langchain-google = ["langchain-google-genai", "langchain-google-vertexai"]
 langchain-nvidia = ["langchain-nvidia-ai-endpoints"]
+colbert = ["colbert-ai", "pyarrow", "torch", "cassio"]
```
nicoloboschi, line 34: cassio can be removed from here, it's always imported

ragstack/colbert/__init__.py
```diff
@@ -0,0 +1,19 @@
...
+    ColbertAstraRetriever,
+    max_similarity_torch,
+    PerTokenEmbeddings,
+    PassageEmbeddings,
+    TokenEmbeddings,
+    DEFAULT_COLBERT_MODEL,
+    DEFAULT_COLBERT_DIM,
+)
+
+__version__ = "0.0.1"
```
nicoloboschi, line 19: no need, please remove

ragstack/colbert/astra_retriever.py
```diff
@@ -0,0 +1,162 @@
...
+        # if the query has fewer than a predefined number of of tokens Nq,
+        # colbertEmbeddings will pad it with BERT special [mast] token up to length Nq.
+        #
+        query_encodings = self.colbertEmbeddings.encode_query(
+            query, query_maxlen=query_maxlen
+        )
+
+        # the min of query_maxlen is 32
+        top_k = max(math.floor(len(query_encodings) / 2), 16)
+        if self.verbose:
```
nicoloboschi, line 105: please remove, let's used logging.debug as verbose log

ragstack/colbert/astra_retriever.py
```diff
@@ -0,0 +1,162 @@
...
+            scores[(title, part)] = sum(
+                max_similarity_torch(qv, embeddings_for_part, self.is_cuda)
+                for qv in query_encodings
+            )
+        # load the source chunk for the top k documents
+        docs_by_score = sorted(scores, key=scores.get, reverse=True)[:k]
+        answers = []
+        rank = 1
+        for title, part in docs_by_score:
+            rs = self.astra.session.execute(
```
nicoloboschi, line 147: here there's room for improvement by doing it async 

then sort answers again before returning 

no need to do it now

ragstack/colbert/cassandra_db.py
```diff
@@ -0,0 +1,183 @@
...
+
+from .token_embedding import PassageEmbeddings
+
+
+def required_cred(cred: str):
+    if cred is None or cred == "":
+        raise ValueError("Please provide credentials")
+
+
+class AstraDB:
```
nicoloboschi, line 15: shoulnd't be Cassandra?

ragstack/colbert/cassandra_db.py
```diff
@@ -0,0 +1,183 @@
...
+        keyspace: str = "colbert128",
+        verbose: bool = False,
+        timeout: int = 60,
+        **kwargs,
+    ):
+        required_cred(secure_connect_bundle)
+        required_cred(astra_token)
+
+        # self.cluster = Cluster(**kwargs)
+        self.cluster = Cluster(
```
nicoloboschi, line 29: still no cassio ? 
I think you can just assume here that cassio.init has been called and get the session by using

`session = cassio.resolve_session()`

.github/workflows/ci-unit-tests.yml
```diff
@@ -43,9 +43,12 @@ jobs:
...
       - name: "Build"
         run: |
-          poetry install --no-root
+          poetry install --no-root -E colbert
           poetry build
 
       - name: Run ragstack-ai unit tests
+        env:
+          COLBERT_ASTRA_TOKEN: ${{ secrets.COLBERT_ASTRA_TOKEN }}
+          COLBERT_ASTRA_SCB: ${{ secrets.COLBERT_ASTRA_SCB }}
```
nicoloboschi, line 52: We cannot rely on an astra database like this in the CI 

we should use cassio, and run the integration tests after the db creation in _run_e2e_tests.yaml file

additionally, to make it easier, we can spin up a docker container with dse. You can copy and paste this class 
https://github.com/datastax/ragstack-ai/blob/main/ragstack-e2e-tests/e2e_tests/test_utils/cassandra_container.py

tests/unit-tests/test_astra_embedding_retrieval.py
```diff
@@ -0,0 +1,112 @@
...
+    for pEmbd in passageEmbeddings:
+
+        print(f"passage embedding title {pEmbd.title()} size {len(pEmbd.get_all_token_embeddings())}")
+        print(f"passagen embedding id {pEmbd.id()}")
+        for tokenEmbd in pEmbd.get_all_token_embeddings():
+            print(f"    token embedding id {tokenEmbd.id} parent {tokenEmbd.parent_id} size {len(tokenEmbd.get_embeddings())}")
+
+
+    # Fetch the Base64 encoded string from the environment variable
+    encoded_zip = os.getenv('COLBERT_ASTRA_SCB')
```
nicoloboschi, line 73: We cannot rely on an astra database like this in the CI

we should use cassio, and run the integration tests after the db creation in _run_e2e_tests.yaml file

additionally, to make it easier, we can spin up a docker container with dse. You can copy and paste this class
https://github.com/datastax/ragstack-ai/blob/main/ragstack-e2e-tests/e2e_tests/test_utils/cassandra_container.py

tests/unit-tests/test_astra_embedding_retrieval.py
```diff
@@ -0,0 +1,112 @@
...
+    output_zip_path = '/tmp/secure-connect-mingv1.zip'
+
+    # Write the binary data to a new file
+    with open(output_zip_path, 'wb') as zip_file:
+        zip_file.write(decoded_zip)
+
+
+    # astra db
+    astra = AstraDB(
+        secure_connect_bundle=output_zip_path,
```
nicoloboschi, line 88: if we use cassio, we don't need any secure bundle

tests/unit-tests/test_astra_embedding_retrieval.py
```diff
@@ -0,0 +1,112 @@
+# test
```
nicoloboschi, line 1: please create a directory for integration tests
`tests/integration-tests`

nicoloboschi (changes requested):

ragstack/colbertbase/colbertbase/astra_db.py
michaelsembwever: It is good practice yes. 

The logical schema rational doesn't apply here. 

C*/DSE/Astra tables are designed by their read patterns.  And tables are intended to be sparse and have multi-dimensional growth, as opposed to SQL's one-dimensional growth.

So in the CQL paradigm this is one table that is storing only one type of information. 

There's performance wins here.  When reading the `bert_embedding_no=x` row the subsequent read of the `bert_embedding_no=-1` is likely going to be from pagecache (bc same partition key and rows close together).

ragstack/colbert/astra_retriever.py
zzzming: adde async execution

.github/workflows/ci-unit-tests.yml
zzzming: The integration test uses a Cassandra container now.

ragstack/colbert/cassandra_db.py
zzzming: The integration test uses a Cassandra container now.

tox.ini
```diff
@@ -1,12 +1,17 @@
 [tox]
 min_version = 4.0
-envlist = py311
+envlist = py310
```
nicoloboschi, line 3: why?

ragstack/colbertbase/colbertbase/astra_db.py
zzzming: switched to a single table

tox.ini
zzzming: rollbacked. I was testing locally

PR #337, opened 2024-03-18
auto configurable query_maxlen

zzzming:
Since the embedding size is determined by query_maxlen, it does not grow as the number tokens increases. So we added a function to increase query_maxlen up to 512 based on query's token.


ragstack/colbert/colbert_embedding.py
```diff
@@ -129,19 +148,30 @@ def encode_queries(
         query: Union[str, List[str]],
         full_length_search: bool = False,
         query_maxlen: int = 32,
+        enable_dynamic_query_maxlen: bool = True,
```
epinzur, line 151: is there anyway to specify this with one parameter? like maybe if `query_maxlen` isn't set, then we use dynamic?

epinzur (approved):

ragstack/colbert/colbert_embedding.py
zzzming: fxied

PR #338, opened 2024-03-19
reconcile title and id in PerPassageEmbeddings

zzzming:
Replace the "title" field in the PerPassageEmbeddings and use "id" stead from the last code review.

### DIFF
--- a/ragstack/colbert/token_embedding.py
+++ b/ragstack/colbert/token_embedding.py
@@ -12,13 +12,11 @@ class PerTokenEmbeddings:
 
     def __init__(
         self,
-        id: int,
-        part: int,
-        parent_id: uuid.UUID = None,
-        title: str = "",
+        id: int,  # id is the embedding id
+        part: int,  # part is the part id of the passage
+        title: str = "",  # title is the title of the passage
     ):
         self.id = id
-        self.parent_id = parent_id
         self.__embeddings = []
         self.title = title
         self.part = part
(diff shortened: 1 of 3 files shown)


ragstack/colbert/token_embedding.py
```diff
@@ -12,13 +12,11 @@ class PerTokenEmbeddings:
 
     def __init__(
         self,
-        id: int,
-        part: int,
-        parent_id: uuid.UUID = None,
-        title: str = "",
+        id: int,  # id is the embedding id
+        part: int,  # part is the part id of the passage
+        title: str = "",  # title is the title of the passage
```
epinzur, line 17: i think we should drop `title` completely from this code. If someone wants to store the name of the file or a document `title`, they will do it in the metadata.

ragstack/colbert/token_embedding.py
```diff
@@ -43,26 +41,32 @@ class PassageEmbeddings:
...
+    __model: str
+    __dim: int
 
     def __init__(
         self,
         text: str,
-        title: str = "",
-        part: int = 0,
-        id: uuid.UUID = None,
+        title: str = "",  # keep this as backward compatibility, use id instead
```
epinzur, line 51: do we need to worry about backward compatibility? this hasn't been released yet, correct?

epinzur (changes requested):

ragstack/colbert/token_embedding.py
nicoloboschi: nope

epinzur:
merged https://github.com/datastax/ragstack-ai/pull/347 instead of this. closing the PR.

PR #340, opened 2024-03-19
colbert: add lint to colbert package and tests

nicoloboschi:


ragstack/colbert/colbert_embedding.py
```diff
@@ -161,50 +158,54 @@ def encode_queries(
...
+        logging.info(
+            f"{len(tokens[0])} tokens in first query with query_maxlen {fixed_length}"
+        )
 
         self.checkpoint.query_tokenizer.query_maxlen = fixed_length
 
         # All query embeddings in the ColBERT documentation
         # this name, EQ or Q, maps the exact name in most colBERT papers
-        Q = self.checkpoint.queryFromText(
+        queriesQ = self.checkpoint.queryFromText(
```
cbornet, line 172:
```suggestion
        queries = self.checkpoint.queryFromText(
```

ragstack/colbert/colbert_embedding.py
```diff
@@ -161,50 +158,54 @@ def encode_queries(
...
             queries, bsize=bsize, to_cpu=True, full_length_search=full_length_search
         )
-
-        return Q
+        return queriesQ
```
cbornet, line 175:
```suggestion
        return queries
```

ragstack/colbert/vector_store.py
```diff
@@ -27,8 +22,28 @@ def put_document(self, document: str, metadata: dict):
...
+    body: str
+    rank: int
+    score: Number
+
+
+class ColBERTVectorStoreRetriever(ABC):
+    @abstractmethod
+    def close(self):
+        """Close the store."""
+        pass
```
cbornet, line 42:
```suggestion
```

ragstack/colbert/vector_store.py
```diff
@@ -27,8 +22,28 @@ def put_document(self, document: str, metadata: dict):
...
+
+    @abstractmethod
+    def retrieve(
+        self, query: str, k: Optional[int], query_maxlen: Optional[int], **kwargs
+    ) -> List[Document]:
+        """Retrieve documents from the store"""
+        pass
```
cbornet, line 49:
```suggestion
```

cbornet:
Weird. Colbert is installed but the test complains that it doesn't find the module 😕

ragstack/colbert/colbert_embedding.py
zzzming: Is there a way to align what's the variable in the paper and doc? I think the code should reflect what the paper describes. If Q cannot used, we should at least use queriesQ or something with a Q

zzzming (approved):

ragstack/colbert/colbert_embedding.py
cbornet: Then just `q` lowercase ?

PR #347, opened 2024-03-21
replaced "title" with "id" for ColBERT document identifier

epinzur:
ColBERT should not require a `title` for each document. It should be an optional `id`, that is generated if not passed.

If a developer wants to save the `filename` or other `title` for a document, they should do it in the metadata.

This PR is in place of https://github.com/datastax/ragstack-ai/pull/338

Also made a bunch of other name changes.

### DIFF
--- a/ragstack/colbert/colbert_embedding.py
+++ b/ragstack/colbert/colbert_embedding.py
@@ -5,7 +5,7 @@ import torch
 from torch import Tensor
 import uuid
 from .token_embedding import TokenEmbeddings, PerTokenEmbeddings, PassageEmbeddings
-from .constant import MAX_MODEL_TOKENS
+from .constant import MAX_MODEL_TOKENS, DEFAULT_COLBERT_MODEL
 
 from colbert.infra import Run, RunConfig, ColBERTConfig
 from colbert.indexing.collection_encoder import CollectionEncoder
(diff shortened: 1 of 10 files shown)


nicoloboschi (approved):

zzzming (approved):
one minor comment. We don't need parent_id in the PerTokenEmbeddings.

ragstack/colbert/colbert_embedding.py
```diff
@@ -123,17 +123,15 @@ def __init__(
             self.checkpoint = self.checkpoint.cuda()
 
     def embed_documents(
```
nicoloboschi, line 125: is "document" correct here? 
if so, `embed_query` should be embed_document

Probably we should use embed_texts and embed_text

nicoloboschi (approved):

ragstack/colbert/colbert_embedding.py
epinzur: this is just for implementing the abstract method in this class: https://github.com/datastax/ragstack-ai/pull/347/files#diff-c553042020c3e53b973244f9a67fb153ece342591f1ed32b563f97ce7f2ff0faR76

Which I'm not sure if we need. What do you think?

nicoloboschi: yes let's keep the interface, code is much leaner to read. Wonder if we want to rename the interface methods too

PR #349, opened 2024-03-22
Add mypy for type checking

cbornet:
Adds mypy and fixes a bunch of type issues.
Not every issue is fixed by this PR yet so atm mypy is not enforced in CI.


epinzur (approved):

nicoloboschi (approved):
Nice

PR #351, opened 2024-03-22
fix the annoying 'consistency level' warning

epinzur:
when running retrieval with ColBERT, I'm currently getting the following warning about 500 times PER QUERY!

```
WARNING:cassandra.protocol:Server warning: Top-K queries can only be run with consistency level ONE / LOCAL_ONE / NODE_LOCAL. Consistency level LOCAL_QUORUM was requested. Downgrading the consistency level to LOCAL_ONE.
```

This tells the DB to use consistency level to LOCAL_ONE on the query, which it will do anyway.

I've confirmed with this change, the warning log line no longer appears.

### DIFF
--- a/ragstack/colbert/cassandra_retriever.py
+++ b/ragstack/colbert/cassandra_retriever.py
@@ -8,6 +8,8 @@ from torch import tensor
 import torch
 import math
 
+from cassandra import ConsistencyLevel
+
 from .vector_store import ColBERTVectorStoreRetriever, Document
 
 # max similarity between a query vector and a list of embeddings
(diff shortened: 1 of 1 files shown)


ragstack/colbert/cassandra_retriever.py
```diff
@@ -113,9 +115,12 @@ def retrieve(
         doc_futures = []
         for qv in query_encodings:
             # per token based retrieval
-            doc_future = self.vector_store.session.execute_async(
-                self.vector_store.query_colbert_ann_stmt, [list(qv), top_k]
-            )
+            bound_statement = self.vector_store.query_colbert_ann_stmt.bind([list(qv), top_k])
+
+            # top-k queries can only be run with consistency level ONE / LOCAL_ONE / NODE_LOCAL
```
zzzming, line 120: I see Nicolo's comment on this might be Astra only. Can you provide an option to allow user to suppress the warning as needed? 

LGTM

zzzming (approved):

ragstack/colbert/cassandra_retriever.py
```diff
@@ -113,9 +115,12 @@ def retrieve(
...
+            bound_statement.consistency_level = ConsistencyLevel.LOCAL_ONE
```
nicoloboschi, line 121: I think it's preferable to fix this problem in CassIO/cassandra driver.

@epinzur for your tests, you can disable the warn logging for that particular logger if you want

For users, we should track this on CassIO probably

nicoloboschi (changes requested):

ragstack/colbert/cassandra_retriever.py
epinzur: Stefano is fixing this in CassIO soon. Per jbellis, seems like LOCAL_ONE is the recommended strategy for all cassandra: https://github.com/CassioML/cassio/issues/35#issuecomment-1841724411

I'll close the PR.

epinzur:
This will be fixed in CassIO, via this issue: https://github.com/CassioML/cassio/issues/35

PR #353, opened 2024-03-25
additional renaming and cleanup of colbert module

epinzur:


ragstack/colbert/token_embedding.py
```diff
@@ -3,70 +3,41 @@
 
 from abc import ABC, abstractmethod
 from typing import List
+from torch import Tensor
 
-class PerTokenEmbeddings:
```
epinzur, line 7: we can use Torch.tensor instead of this class

ragstack/colbert/colbert_embedding.py
```diff
@@ -102,21 +101,20 @@ def __init__(
         if self.__cuda:
             self.checkpoint = self.checkpoint.cuda()
 
-    def embed_documents(
-        self, texts: List[str], doc_id: str = ""
-    ) -> List[PassageEmbeddings]:
+    def embed_chunks(self, chunk_texts: List[str], doc_id: str = None) -> List[EmbeddedChunk]:
```
nicoloboschi, line 104:
```suggestion
    def embed_chunks(self, chunk_texts: List[str], doc_id: Optional[str] = None) -> List[EmbeddedChunk]:
```

ragstack/colbert/vector_store.py
```diff
@@ -6,18 +6,16 @@
...
 
     @abstractmethod
-    def close(self):
-        """Close the store."""
-        pass
-
-    @abstractmethod
-    def put_document(self, document: str, metadata: dict):
+    def put_chunks(
+        self, chunks: List[EmbeddedChunk], delete_existing: bool = False
```
nicoloboschi, line 17:
```suggestion
        self, chunks: List[EmbeddedChunk], delete_existing: Optional[bool] = False
```

nicoloboschi (changes requested):

PR #356, opened 2024-03-28
cassandra query timeout in colBERT retriever

zzzming:


ragstack/colbert/cassandra_store.py
```diff
@@ -13,11 +13,15 @@ class CassandraColbertVectorStore(ColbertVectorStore):
...
-    def __init__(self, session: Session, keyspace: str, table_name: str):
+    def __init__(
+            self, session: Session, keyspace: str, table_name: str, timeout: int = 180
+        ):
         self.keyspace = keyspace
         self.session = session
         self.table_name = table_name
         self.full_table_name = f"{keyspace}.{table_name}"
+
+        self.session.default_timeout = timeout
```
epinzur, line 24: if we have this one, do we need the others above?

zzzming: This is a global timeout in the Cassandra. The other is per session execution call. I believe there are benefits to allow a function call to overwrite the global timeout.

PR #360, opened 2024-04-02
colbert config and embedding timeout

zzzming:


ragstack/colbert/colbert_embedding.py
```diff
@@ -54,11 +54,20 @@ class ColbertTokenEmbeddings(TokenEmbeddings):
     encoder: CollectionEncoder
     query_tokenizer: QueryTokenizer
 
+    '''
+    checkpoint is the where the ColBERT model can be specified or downloaded from huggingface
+    doc_maxlen is the number tokens each passages are truncated to
```
epinzur, line 59: `doc_maxlen` is the number tokens each passage is truncated to

ragstack/colbert/runner.py
```diff
@@ -58,31 +58,49 @@ def __init__(self, nranks: int = 1):
...
+            if p.is_alive():
+                logging.warn(f"embedding process timed out PID: {info[0]}, Name: {info[1]}")
+                timed_out_processes.append(p)
+            else:
+                logging.info(f"joined embedding process ID: {info[0]}, Name: {info[1]}")
+
+        if timed_out_processes:
+            raise Exception("One or more processes did not complete within the timeout period.")
+        else:
+            logging.info("all processes joined successfully.")
```
epinzur, line 99: maybe "all processes completed" ?

epinzur: and maybe

ragstack/colbert/colbert_embedding.py
```diff
@@ -54,11 +54,20 @@ class ColbertTokenEmbeddings(TokenEmbeddings):
     encoder: CollectionEncoder
     query_tokenizer: QueryTokenizer
 
+    '''
+    checkpoint is the where the ColBERT model can be specified or downloaded from huggingface
```
epinzur, line 58: is there a better name than `checkpoint`? maybe `colbert_model_url`?

epinzur (approved):

ragstack/colbert/colbert_embedding.py
zzzming: The reason uses checkpoint is because the ColBERT library uses the same name. I added another argument called colbert_model_url that can overwrite `checkpoint` if it exists. Is that ok?

PR #361, opened 2024-04-02
created docstrings for the root colBERT package

epinzur:
these are initial docstrings, created with the help of chatGPT.  We can revise these on another PR.

also made a few parameter name changes, like continuing to rename `passages` to `chunks`


ragstack/colbert/chunk_encoder.py
```diff
@@ -0,0 +1,147 @@
+"""
```
epinzur, line 1: new version of `passage_encoder.py`

pyproject.toml
```diff
@@ -27,6 +27,8 @@ langchain-nvidia-ai-endpoints = { version = "0.0.3", optional = true }
 colbert-ai = { version = "0.2.19", optional = true }
 pyarrow = { version = "14.0.1", optional = true }
 torch = { version = "2.2.1", optional = true }
+faiss-gpu = { version = ">=1.7.0", optional = true }
```
zzzming, line 30: Can we just lock on a version? Faiss can have compatibility issues if it's not tested.

zzzming (approved):
I don't think we need faiss as we discussed

pyproject.toml
epinzur: i just removed it from the dependency list

PR #362, opened 2024-04-03
made various changes to cleanup retrieval

epinzur:
made a bunch of changes that generally make sure our ColBERT retrieval code interacts nicely with Cassandra (doesn't overload the database when making queries). 

There is little change in retrieval latency with these changes running on CPU.

* wrapped async calls to cassandra inside a `ThreadPoolExecutor`.  
  * We should experiment to see what value of `max_casandra_workers` works the best.

* automatically dropping down to half-precision when calculating chunk scores, if running on GPU and the GPU supports it.  
  * We should test to see if `is_fp16` flag is getting set to `True` on boxes with CUDA support. 
  * I "think" that half-precision won't hurt our results, and chatGPT claims this can "significantly speed up our calculations & reduce memory usage on the GPU".

### DIFF
--- a/ragstack/colbert/cassandra_retriever.py
+++ b/ragstack/colbert/cassandra_retriever.py
@@ -209,7 +209,7 @@ class ColbertCassandraRetriever(ColbertVectorStoreRetriever):
         ) -> Tuple[Tuple[str, int], List[Tensor]] | None:
             try:
                 async_future = self.vector_store.session.execute_async(
-                    self.vector_store.query_chunk_stmt,
+                    self.vector_store.query_colbert_chunks_stmt,
                     [doc_id, chunk_id],
                     timeout=query_timeout,
                 )
--- a/ragstack/colbert/cassandra_retriever.py
+++ b/ragstack/colbert/cassandra_retriever.py
@@ -10,10 +10,11 @@ variety of hardware environments.
 
 import logging
 import math
+from concurrent.futures import ThreadPoolExecutor, as_completed
 from typing import Any, Dict, List, Optional, Set, Tuple
 
 import torch
-from cassandra.cluster import ResponseFuture
+from cassandra import ReadTimeout
 from torch import Tensor
 
 from .cassandra_store import CassandraColbertVectorStore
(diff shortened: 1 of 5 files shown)


ragstack/colbert/cassandra_retriever.py
```diff
@@ -69,6 +106,9 @@ class ColbertCassandraRetriever(ColbertVectorStoreRetriever):
                                                     Cassandra database.
         colbert_embeddings (ColbertTokenEmbeddings): The ColbertTokenEmbeddings instance for encoding queries.
         is_cuda (bool): A flag indicating whether to use CUDA (GPU) for computation.
+        is_fp16 (bool): A flag indicating whether to half-precision floating point operations on CUDA (GPU).
+                        Has no effect on CPU computation.
+        max_casandra_workers: The maximum number of concurrent requests to make to Cassandra on a per-retrieval basis.
```
nicoloboschi, line 111:
```suggestion
        max_workers: The maximum number of concurrent requests to make to the vector store on a per-retrieval basis.
```

ragstack/colbert/cassandra_retriever.py
```diff
@@ -77,7 +117,9 @@ class ColbertCassandraRetriever(ColbertVectorStoreRetriever):
 
     vector_store: CassandraColbertVectorStore
     colbert_embeddings: ColbertTokenEmbeddings
+    max_casandra_workers: int
```
nicoloboschi, line 120:
```suggestion
    max_workers: int
```

and default?

ragstack/colbert/cassandra_retriever.py
```diff
@@ -94,17 +137,181 @@ def __init__(
...
+        executor_futures = {
+            self.executor.submit(query_and_process, qv): qv for qv in query_encodings
+        }
+        for executor_future in as_completed(executor_futures):
+            query_vector = executor_futures[executor_future]
+            try:
+                result_chunks = executor_future.result()
+                chunks.update(result_chunks)
+            except Exception as exc:
+                logging.error(
```
nicoloboschi, line 193: should we raise the exception outside here?

ragstack/colbert/cassandra_retriever.py
```diff
@@ -94,17 +137,181 @@ def __init__(
...
+                    self.vector_store.query_colbert_chunks_stmt,
+                    [doc_id, chunk_id],
+                    timeout=query_timeout,
+                )
+                rows = async_future.result()
+                return (doc_id, chunk_id), [
+                    torch.tensor(row.bert_embedding) for row in rows
+                ]
+            except ReadTimeout:
+                logging.warn(f"Query timeout for doc_id {doc_id}, chunk_id {chunk_id}")
```
nicoloboschi, line 221: shall we retry? 
In general I believe if one query fails, we should raise the error to the caller

nicoloboschi (changes requested):
LGTM in general

I think we should improve the error handling

ragstack/colbert/cassandra_retriever.py
epinzur: we def need better error handling throughout this package.  I'd prefer to focus on that in another PR. 

at least this PR starts to set up a bit of it.

ragstack/colbert/cassandra_retriever.py
epinzur: i'll change this to error level for now.

ragstack/colbert/cassandra_retriever.py
```diff
@@ -86,6 +128,7 @@ def __init__(
         self,
         vector_store: CassandraColbertVectorStore,
         colbert_embeddings: ColbertTokenEmbeddings,
+        max_casandra_workers: Optional[int] = 10,
```
nicoloboschi, line 131: please correct the typo

PR #368, opened 2024-04-10
reduce paddings on query_maxlen in the default setting

zzzming:
reduce paddings on query_maxlen in the default setting although user can still specify the query_maxlen

### DIFF
--- a/ragstack/colbert/colbert_embedding.py
+++ b/ragstack/colbert/colbert_embedding.py
@@ -25,16 +25,12 @@ from .runner import Runner
 from .token_embedding import EmbeddedChunk, TokenEmbeddings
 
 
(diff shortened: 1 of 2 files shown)


ragstack/colbert/colbert_embedding.py
```diff
@@ -25,33 +25,24 @@
 from .token_embedding import EmbeddedChunk, TokenEmbeddings
 
 
-def calculate_query_maxlen(tokens: List[List[str]], min_num: int, max_num: int) -> int:
+def calculate_query_maxlen(tokens: List[List[str]]) -> int:
```
epinzur, line 28: also need to update the call where this is used:

https://github.com/datastax/ragstack-ai/blob/main/ragstack/colbert/colbert_embedding.py#L224

PR #370, opened 2024-04-10
[CI]: enable gpt4 tests

jordanrfrazier:
Re-enables GPT4 tests with lowered test frequency to reduce costs.


ragstack-e2e-tests/e2e_tests/langchain/test_compatibility_rag.py
```diff
@@ -246,8 +246,8 @@ def nvidia_aifoundation_mixtral8x7b_llm():
     "embedding,llm",
     [
         ("openai_ada002_embedding", "openai_gpt35turbo_llm"),
-        # ("openai_3small_embedding", "openai_gpt4_llm"),
-        # ("openai_3large_embedding", "openai_gpt4_llm_streaming"),
+        ("openai_3small_embedding", "openai_gpt4_llm"),
```
nicoloboschi, line 249: maybe we can use gpt35turbo for the streaming + 3large? 
in this way we still test gpt4 but it's cheaper


also I think we should update llama index tests as well

nicoloboschi (approved):

PR #371, opened 2024-04-10
[tests]: change fixtures to functions

jordanrfrazier:
This change postpones the initialization of the llms and embedding objects. This allows us to do things like check the config for nemo guardrails to skip the test before attempting init. 

Previously, if the init failed, we weren't correctly skipping the `nemo` test, meaning they were displayed as failing within testspace, rather than automatically skipping them as expected.


nicoloboschi (approved):
Good work

PR #387, opened 2024-04-22
colbert naming and organizational chagnes

epinzur:
This is a proposed change to the colbert naming scheme before we do the initial release.

I've disliked the long class names we've had like:

```
from ragstack-colbert import CassandraColbertVectorStore, ColbertCassandraRetriever
```

This simplifies to make the Cassandra part assumed (since it is a datastax product)
```
from ragstack-colbert import ColbertVectorStore, ColbertRetriever
```

But I could see the benefit in keeping the more descriptive names too.


zzzming (changes requested):
I don't like long names either. It looks good, but I think we need keep the `Cassandra` part. One option is to create a folder called cassandra. So that we can still call it the new name ColBERTVectorStore.

epinzur:
new names are:

```
ColbertRetriever
CassandraVectorStore

ColbertLCRetriever
ColbertLIRetriever
```

libs/colbert/ragstack_colbert/base_retriever.py
```diff
@@ -0,0 +1,48 @@
+"""
```
epinzur, line 1: changed all the interfaces to be prefixed with `base_` and renamed the class name too.

libs/colbert/ragstack_colbert/base_embedding.py
```diff
@@ -10,7 +10,7 @@
 from .chunks import EmbeddedChunk
 
 
-class TokenEmbeddings(ABC):
+class BaseEmbedding(ABC):
```
zzzming, line 13: So the class name should not be plural right?

zzzming (approved):

nicoloboschi (approved):

libs/colbert/ragstack_colbert/base_embedding.py
epinzur: i think so... i keep going back and forth on this

libs/colbert/ragstack_colbert/cassandra_vector_store.py
```diff
@@ -1,5 +1,5 @@
 """
-This module provides an implementation of the ColbertVectorStore abstract base class, specifically designed
+This module provides an implementation of the BaseVectorStore abstract class, specifically designed
 for use with a Cassandra database backend. It allows for the efficient storage and management of text embeddings
```
michaelsembwever, line 3: nit: the technology is named Apache Cassandra, and should initially be referenced as such.  subsequent usages can them be shorten to just Cassandra.

this is mostly FYI.

whether you want to do this here, or somewhere like in the README where a user would first come across it, i leave up to you.

michaelsembwever:
nice improvement! makes the separation of technologies it's using clearer.

the comment was made (elsewhere) that… 

> Cassandra part assumed (since it is a datastax product)

while this is nit-picking and most likely taking the meaning of that sentence out of content, it's worth the gentle reminder that Cassandra is not a datastax product, and any indication of such is a legal branding violation. (The product is Apache Cassandra, and it belongs to the ASF.)

PR #396, opened 2024-04-24
colbert integration test on cassandra

zzzming:
* use a nested event loop instead of creating a new event loop. Some application already creates an event loop (i.e. colab)
* fix a file name typo
* reinstate c* docker based integration test


epinzur:
👍

PR #398, opened 2024-04-27
use query token mask to optimize for effective tokens

zzzming:
Revert back to use the standard query_maxlen for query token embeddings, this is the library's default.

Instead we use the query embedding's tensorize()'s mask to identify effective query tokens. It excludes 101, 1, 102, and 103.

### DIFF
--- a/libs/colbert/ragstack_colbert/base_embedding_model.py
+++ b/libs/colbert/ragstack_colbert/base_embedding_model.py
@@ -62,3 +62,21 @@ class BaseEmbeddingModel(ABC):
         Returns:
             Tensor: A tensor representing the embedded query.
         """
+
+    @abstractmethod
+    def optimized_query_embeddings(
+        self, query: str, query_maxlen: int = -1
+    ) -> Tensor:
+        """
+        Embeds a single query text into its vector representation for optimized retrieval.
+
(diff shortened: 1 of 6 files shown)


libs/colbert/ragstack_colbert/base_embedding_model.py
```diff
@@ -62,3 +62,21 @@ def embed_query(
         Returns:
             Tensor: A tensor representing the embedded query.
         """
+
+    @abstractmethod
+    def optimized_query_embeddings(
```
epinzur, line 67: seems odd to put this into the interface... we should figure out which method works better and just make it the default for `embed_query` above.

libs/colbert/ragstack_colbert/base_retriever.py
```diff
@@ -27,7 +27,7 @@ def retrieve(
         self,
         query: str,
         k: Optional[int] = None,
-        query_maxlen: Optional[int] = None,
+        query_maxlen: int = 64,
```
epinzur, line 30: i think we have proven that dynamic query_maxlen is superior to using the default of 64. I'd undo this change.

epinzur:
I'm running a test to see if this improves performance. please do not merge until we have data to back up this change.

PR #423, opened 2024-05-14
colbert: python 3.12 support

zzzming:


nicoloboschi (changes requested):
unstructured doesn't support py 3.12 yet 
https://pypi.org/project/unstructured/

nicoloboschi (approved):
added CI and now unstructured is upgraded

PR #447, opened 2024-05-24
feat: knowlege-store concurrent traversal

bjchambers:
This should speed up traversals by issuing the queries for nodes/edges/ids concurrently.


nicoloboschi (approved):
LGTM

PR #450, opened 2024-05-30
ref: Separate out edge extractors

bjchambers:
This is a step towards making the edge extraction extensible. Rather than specifying which extractions to apply with a bunch of hard-coded booleans, this takes a list of extractors.

This has some limitations:
- Extractors use hard-coded columns in the CQL schema. This means that we can't use multiple extractors of the same type, nor can we add new extractors without adjusting the schema. This may be addressed in the future.
- Concurrent execution is aligned between extractors. That is, all tasks from one extractor complete before the next begins. This is likely better than requiring some coordination between extractors, especially if we wish to make this a more generalizable implementation. This can be revisited if it becomes a problem, although `async` versions are likely the best solution.

Additionally, future improvements to be made include:
- Allowing extractors to (configurably) annotate the edges the add
- Moving queries into the extractors and/or generally cleaning up how the extractors interact with the store.


libs/knowledge-store/ragstack_knowledge_store/directed_edge_extractor.py
```diff
@@ -0,0 +1,98 @@
...
+    While `UndirectedEdgeExtractor` links nodes in both directions if they share
+    a keyword, this only creates links from nodes with a "source" to nodes with
+    a matching "target". For example, uses may be the `href` of `a` tags in the
+    chunk and definitions may be the URLs that the chunk is accessible at.
+
+    This may also be used for other forms of references, such as Wikipedia article IDs, etc.
+    """
+
+    def __init__(self,
+                 sources_field: str,
```
bjchambers, line 19: This only partly makes these fields configurable -- note the `KnowledgeStore` is currently still responsible for writing metadata from these fields, so it's not perfect yet. 

Ideally, either the `KnowledgeStore` would consult the extractors (to figure out what columns to populate an how) which could get tricky / tightly coupled or each extractor would be responsible for it's own writes (which could be annoying since there'd be `K+1` writes for each node with `K` extractors).

kerinin (commented):
Ultimately, I would prefer to be able to provide the extraction logic itself, rather than just a mapping from metadata to edges. IE, an extractor that takes a document with no metadata and adds edges based on keywords in the `page_content`, or URLs, etc.

bjchambers:
> Ultimately, I would prefer to be able to provide the extraction logic itself, rather than just a mapping from metadata to edges. IE, an extractor that takes a document with no metadata and adds edges based on keywords in the `page_content`, or URLs, etc.

I mean -- that's possible. The edge extractor function takes the `text` and `metadata`, so it can do whatever it wants? I'm not sure what's missing for you to be able to implement such an extractor.

kerinin:
> > Ultimately, I would prefer to be able to provide the extraction logic itself, rather than just a mapping from metadata to edges. IE, an extractor that takes a document with no metadata and adds edges based on keywords in the `page_content`, or URLs, etc.
> 
> I mean -- that's possible. The edge extractor function takes the `text` and `metadata`, so it can do whatever it wants? I'm not sure what's missing for you to be able to implement such an extractor.

I just meant we should provide these extractors in the future.

bjchambers:
> > > Ultimately, I would prefer to be able to provide the extraction logic itself, rather than just a mapping from metadata to edges. IE, an extractor that takes a document with no metadata and adds edges based on keywords in the `page_content`, or URLs, etc.
> > 
> > 
> > I mean -- that's possible. The edge extractor function takes the `text` and `metadata`, so it can do whatever it wants? I'm not sure what's missing for you to be able to implement such an extractor.
> 
> I just meant we should provide these extractors in the future.

Makes sense. And actually, this should be pretty easy. I think we could just add a variant of the "directed" edge extractor that took a list of ids. This would also let us add edges faster. For instance, if we were doing `hrefs -> url` and new that the URL was the ID, we wouldn't need to look up nodes with the given `url`, allowing this to be a write-only edge extraction. I'll probably add that since it may simplify some experiments with Wikipedia stuff.

libs/knowledge-store/ragstack_knowledge_store/edge_extractor.py
```diff
@@ -0,0 +1,49 @@
...
+    @property
+    def kind(self) -> str:
+        """Return the kind of edge extracted by this."""
+        ...
+
+    def tags(self, text: str, metadata: Dict[str, Any]) -> Set[str]:
+        """Return the set of tags to add for this extraction."""
+        return set()
+
+    def extract_edges(
```
cbornet, line 17: Annotate with `@abstractmethod`

libs/knowledge-store/ragstack_knowledge_store/edge_extractor.py
```diff
@@ -0,0 +1,49 @@
...
+
+        Parameters:
+        - store: KnowledgeStore edges are being extracted for.
+        - texts: The texts of the nodes to be processed.
+        - metadatas: The metadatas of the nodes to be processed.
+
+        Returns:
+        Number of edges extracted involving the given nodes.
+        """
+        ...
```
cbornet, line 32:
```suggestion
```

libs/knowledge-store/ragstack_knowledge_store/edge_extractor.py
```diff
@@ -0,0 +1,49 @@
...
+
+        Parameters:
+        - store: KnowledgeStore edges are being extracted for.
+        - texts: The texts of the nodes to be processed.
+        - metadatas: The metadatas of the nodes to be processed.
+
+        Returns:
+        Number of edges extracted involving the given nodes.
+        """
+        return self.extract_edges(store, texts, metadatas)
```
cbornet, line 49: Should we wrap in `run_in_executor` ?

libs/knowledge-store/ragstack_knowledge_store/edge_extractor.py
```diff
@@ -0,0 +1,49 @@
...
+from typing import Any, Dict, Iterable, Set
+
+from knowledge_store import KnowledgeStore
+
+
+class EdgeExtractor(ABC):
+    @property
+    def kind(self) -> str:
+        """Return the kind of edge extracted by this."""
+        ...
```
cbornet, line 11:
```suggestion
```

libs/knowledge-store/ragstack_knowledge_store/edge_extractor.py
```diff
@@ -0,0 +1,49 @@
+from abc import ABC
+from typing import Any, Dict, Iterable, Set
+
+from knowledge_store import KnowledgeStore
+
+
+class EdgeExtractor(ABC):
+    @property
+    def kind(self) -> str:
```
cbornet, line 9: Annotate with `@abstractmethod`

libs/knowledge-store/ragstack_knowledge_store/directed_edge_extractor.py
```diff
@@ -0,0 +1,130 @@
...
+    While `UndirectedEdgeExtractor` links nodes in both directions if they share
+    a keyword, this only creates links from nodes with a "source" to nodes with
+    a matching "target". For example, uses may be the `href` of `a` tags in the
+    chunk and definitions may be the URLs that the chunk is accessible at.
+
+    This may also be used for other forms of references, such as Wikipedia article IDs, etc.
+    """
+
+    def __init__(self, sources_field: str, targets_field: str, kind: str) -> None:
+        """Create a new DirectedEdgeExtractor.
```
cbornet, line 21: I propose to merge the init docstring with the class docstring and only put it on init.
IDEs and online doc builders will use the init docstring for the class if the class doesn't have one.

libs/knowledge-store/ragstack_knowledge_store/directed_edge_extractor.py
```diff
@@ -0,0 +1,130 @@
...
+
+        Params:
+        - sources_field: The metadata field to read sources from.
```
cbornet, line 24: Use google-style `Args`

libs/knowledge-store/ragstack_knowledge_store/edge_extractor.py
```diff
@@ -0,0 +1,49 @@
...
+        return set()
+
+    def extract_edges(
+        self, store: KnowledgeStore, texts: Iterable[str], metadatas: Iterable[Dict[str, Any]]
+    ) -> int:
+        """Add edges for the given nodes.
+
+        The nodes have already been persisted.
+
+        Parameters:
```
cbornet, line 24: Use google-style `Args`

libs/knowledge-store/ragstack_knowledge_store/edge_extractor.py
bjchambers: Probably a good idea. My preference would be for us to truly implement these, but I'll wrap for now.

libs/knowledge-store/ragstack_knowledge_store/directed_edge_extractor.py
```diff
@@ -0,0 +1,127 @@
...
+        a matching "target". For example, uses may be the `href` of `a` tags in the
+        chunk and definitions may be the URLs that the chunk is accessible at.
+
+        This may also be used for other forms of references, such as Wikipedia
+        article IDs, etc.
+
+        Args:
+        - sources_field: The metadata field to read sources from.
+        - targets_field: The metadata field to read targets from.
+        - kind: The kind label to apply to created edges. Must be unique.
```
cbornet:
```suggestion
            sources_field: The metadata field to read sources from.
            targets_field: The metadata field to read targets from.
            kind: The kind label to apply to created edges. Must be unique.
```

libs/knowledge-store/ragstack_knowledge_store/edge_extractor.py
```diff
@@ -0,0 +1,57 @@
...
+        self, store: KnowledgeStore, texts: Iterable[str], metadatas: Iterable[Dict[str, Any]]
+    ) -> int:
+        """Add edges for the given nodes.
+
+        The nodes have already been persisted.
+
+        Args:
+        - store: KnowledgeStore edges are being extracted for.
+        - texts: The texts of the nodes to be processed.
+        - metadatas: The metadatas of the nodes to be processed.
```
cbornet, line 31:
```suggestion
            store: KnowledgeStore edges are being extracted for.
            texts: The texts of the nodes to be processed.
            metadatas: The metadatas of the nodes to be processed.
```

libs/knowledge-store/ragstack_knowledge_store/edge_extractor.py
```diff
@@ -0,0 +1,57 @@
...
+
+        Returns:
+        Number of edges extracted involving the given nodes.
```
cbornet, line 34:
```suggestion
            Number of edges extracted involving the given nodes.
```

libs/knowledge-store/ragstack_knowledge_store/edge_extractor.py
```diff
@@ -0,0 +1,57 @@
...
+        self, store: KnowledgeStore, texts: Iterable[str], metadatas: Iterable[Dict[str, Any]]
+    ) -> int:
+        """Add edges for the given nodes.
+
+        The nodes have already been persisted.
+
+        Parameters:
+        - store: KnowledgeStore edges are being extracted for.
+        - texts: The texts of the nodes to be processed.
+        - metadatas: The metadatas of the nodes to be processed.
```
cbornet:
```suggestion
        Args:
            store: KnowledgeStore edges are being extracted for.
            texts: The texts of the nodes to be processed.
            metadatas: The metadatas of the nodes to be processed.
```

libs/knowledge-store/ragstack_knowledge_store/explicit_edge_extractor.py
```diff
@@ -0,0 +1,55 @@
...
+        Note: This extractor does not check whether the target ID exists. Edges
+        will be created even if the target does not exist. This means traversals
+        over graphs using this extractor may discover nodes that do not exist.
+        Such "phantom IDs" will be filtered out when loading content for the
+        nodes.
+
+        Params:
+        - edges_field: The metadata field containing the IDs of nodes to link to.
+        - kind: The `kind` to apply to edges created by this extractor.
+        - bidir: If true, creates edges in both directions.
```
cbornet:
```suggestion
        Args:
            edges_field: The metadata field containing the IDs of nodes to link to.
            kind: The `kind` to apply to edges created by this extractor.
            bidir: If true, creates edges in both directions.
```

libs/knowledge-store/ragstack_knowledge_store/parent_edge_extractor.py
```diff
@@ -0,0 +1,37 @@
...
+class ParentEdgeExtractor(EdgeExtractor):
+    def __init__(self, parent_field: str = "parent_content_id") -> None:
+        """Extract an edge from a node to it's parent.
+
+        If no parent is defined, no edge will be created.
+        An edge will be created if the `parent_field` is specified, whether or
+        not the parent node exists.
+
+        Params:
+        - parent_field: The metadata field containing the parent content ID.
```
cbornet:
```suggestion
        Args:
            parent_field: The metadata field containing the parent content ID.
```

libs/knowledge-store/ragstack_knowledge_store/undirected_edge_extractor.py
```diff
@@ -0,0 +1,87 @@
...
+    def __init__(
+        self,
+        keyword_field: str = "keywords",
+        kind: str = "kw",
+    ) -> None:
+        """Extract undirected edges between nodes with common keywords.
+
+        Params:
+        - keyword_field: The metadata field to read keywords from.
+        - kind: The kind label to apply to created edges. Must be unique.
```
cbornet:
```suggestion
        Args:
            keyword_field: The metadata field to read keywords from.
            kind: The kind label to apply to created edges. Must be unique.
```

cbornet (approved):

PR #452, opened 2024-05-30
Extract API for coarse knowledge graph

cbornet:


libs/knowledge-store/ragstack_knowledge_store/document_graph.py
```diff
@@ -0,0 +1,114 @@
...
+from langchain_core.callbacks import CallbackManagerForRetrieverRun
+from langchain_core.documents import Document
+from langchain_core.retrievers import BaseRetriever
+from pydantic import Field
+
+
+def _texts_to_documents(
+    texts: Iterator[str],
+    metadatas: Iterator[dict],
+) -> Iterator[Document]:
```
bjchambers, line 16: We may want ot consider how we can support diferent kindso f nodes -- especially with things like images and such. We may want a `TextNode` or something like that.

Also, rather than going texts->Document should we introduce the `DocumentNode` or somethnig you had mentioned?

libs/knowledge-store/ragstack_knowledge_store/document_graph.py
```diff
@@ -0,0 +1,114 @@
...
+        text,
+        metadata,
+    ) in zip(texts, metadatas):
+        yield Document(
+            page_content=text,
+            metadata=metadata,
+        )
+
+
+class DocumentGraph(ABC):
```
bjchambers, line 27: Comments, etc.

libs/knowledge-store/ragstack_knowledge_store/document_graph.py
```diff
@@ -0,0 +1,114 @@
...
+        documents = _texts_to_documents(iter(texts), metadatas_it)
+        return self.add_documents(documents, ids=ids, links=links, **kwargs)
+
+    @abstractmethod
+    def add_documents(
+        self,
+        documents: Iterable[Document] = None,
+        *,
+        ids: Optional[Iterable[str]] = None,
+        links: Iterable[Tuple[str, str]] = None,
```
bjchambers, line 48: I don't know that these will be enough to create links. Can you elaborate on how this would work?

libs/knowledge-store/ragstack_knowledge_store/document_graph.py
```diff
@@ -0,0 +1,114 @@
...
+    def as_retriever(
+        self, retrieval_kwargs: dict[str, Any], **kwargs: Any
+    ) -> "DocumentGraphRetriever":
+        """Return a Retriever for retrieving from this knowledge store..
+
+        The initial nodes are retrieved using a search.
+        Additional nodes are discovered up to the given depth from those starting nodes.
+
+        Args:
+            retrieval_kwargs: Keyword arguments to pass to the retrieval function.
```
bjchambers, line 86: I don't love this pattern -- but maybe with the `mypy` overload we can at least document the "standard" set of arguments?

cbornet: This is the pattern used for the VectorStoreRetriever.
It's done like this since the search function and its attributes can vary base on the search_type param.

libs/knowledge-store/ragstack_knowledge_store/document_graph.py
```diff
@@ -0,0 +1,114 @@
...
+        documents: Iterable[Document] = None,
+        *,
+        ids: Optional[Iterable[str]] = None,
+        links: Iterable[Tuple[str, str]] = None,
+        **kwargs: Any,
+    ) -> List[str]:
+        """Add documents to the graph"""
+
+    @abstractmethod
+    def traversing_retrieve(
```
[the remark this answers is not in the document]
cbornet: I don't know if it's better to have one uber function vs multiple functions...

libs/knowledge-store/ragstack_knowledge_store/document_graph.py
cbornet: The tuple is (from_id, to_id). This is like this for simplicity but maybe adding a `Link` class that can be extended would be better.

libs/knowledge-store/ragstack_knowledge_store/document_graph.py
cbornet: I removed DocumentNode because it was close to Document an I felt it didn't add much value. I can add it back.

bjchambers: I wonder if we want to have a separate `Node` and then have `DocumentNode` basically be a `Document + Node`. Specifically, this would allow us to introduce `ImageNode`, etc. in the future?

libs/knowledge-store/ragstack_knowledge_store/document_graph.py
bjchambers: There's a few difficulties with making that form useful.

1. IDs may not have been assigned
2. Different kind of link inferences may rely on things being written first / DB queries / etc. I'm not sure it would be possible to infer all of the links ahead of time (may need to compute embeddings, etc.).

I wonder if we could make this use #450 -- basically, we could have an edge extractor that looked for explicit links in the metadata. This would be a list of pairs like this. Then, the user could choose to have that edge extractor or not?

libs/knowledge-store/ragstack_knowledge_store/document_graph.py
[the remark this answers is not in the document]
bjchambers: Yeah -- I find the `@overload` pattern a pain -- I think separate named functions can be nice. I guess my thoughts here are:

1. Multiple functions: Probably nicest from a documentation perspective. Easy to see the docs for a specific function, how different things work, etc. May lead to duplicity in implementation, but not horrible. Implementations can just have the abstract / unimplemented behavior, so not much burden.
2. One uber / overloaded function: Can use the `@overload` for types, but documentation gets messy. Main benefit is if we want to make it really clear "there are N ways to retrieve, and here are your choices". But that can be accomplished with references in the docs, so I think I prefer multiple functions...
3. One function with a bunch of hidden `**kwargs`. This is my least favorite. It's really hard to figure out which kwargs to use for different things, etc.

So -- I guess this comes back to separate methods. Thoughts on naming to show relationships? I'm wondering if it would be better to have `retrieve_traverse` (eg., use `retrieve_` as a prefix to group them?)

libs/knowledge-store/ragstack_knowledge_store/document_graph.py
bjchambers: That makes sense. We can do the `@overload` to convey what the attributes are for a specific search type.

(although, we could call this `as_retriever_traverse` or `as_traversing_retriever`?

But, if we want `search_type`:

```
@overload
def as_retriever(self, search_type: Literal["traverse"], *, k: int = 4, depth: int = 1, **kwargs): ...

@overload
def as_retriever(self, search_type: Literal["mmr"], *, n: int = 10, max_depth: int = 2, **kwargs): ...

def as_retriever(self, search_type: str, **kwargs):
   ...
```

This let's us have types (and maybe docs) for specific search types, will still preserving the pattern from `VectorStoreRetriever` of allowing attributes to vary based on search_type.

libs/knowledge-store/ragstack_knowledge_store/document_graph.py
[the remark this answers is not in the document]
kerinin: Do we actually support different retrieval methods? It seems like the point of using this store is to do hybrid graph/vector retrieval, and between `k` and `depth` you can control how that retrieval behaves. Isn't `traversing_retrieve(k=3, depth-0) ~= retrieve(k=3)`?

libs/knowledge-store/ragstack_knowledge_store/document_graph.py
kerinin: If we're going to keep the `links` parameter, it's probably worth creating an `Edge` class to represent this tuple, just so it's clear what the schema is and we have a place to document things.

That being said, I think I would prefer to hide the edges and have them be created by the store based on a configured set of extractors (I think this is #450?)

libs/knowledge-store/ragstack_knowledge_store/document_graph.py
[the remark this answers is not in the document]
bjchambers: I mean, we plan to right?

libs/knowledge-store/ragstack_knowledge_store/document_graph.py
cbornet: This proposal is compatible with #450 as edge extractors are configured in the implementation init and we pass the metadata in the `add_documents`/`add_texts` methods.
Ok to create `Link` or `Edge` rather than `Tuple[str, str]`.
The idea here is to support edges/relationships created outside of the DB. For instance, from a process that would build the full graph externally and then dump it to the DB. This would have the advantage of being DB independent.

libs/knowledge-store/ragstack_knowledge_store/document_graph.py
cbornet: For now, removing `retrieval_kwargs`.

libs/knowledge-store/ragstack_knowledge_store/document_graph.py
cbornet: You mean we have `add_nodes(nodes: Nodes)` and implementations do type checking to handle the different Node types ? Raising exceptions when they receive a Node type they don't support ?

libs/knowledge-store/ragstack_knowledge_store/document_graph.py
cbornet: Yes, I will add doc everywhere once the API is agreed.

libs/knowledge-store/ragstack_knowledge_store/document_graph.py
cbornet: done.PTAL.

libs/knowledge-store/ragstack_knowledge_store/document_graph.py
cbornet: Edge extractors are very tied to Cassandra, so I don't think it's possible to pull them to the main API atm.

libs/knowledge-store/ragstack_knowledge_store/document_graph.py
cbornet: Done. `as_retriever` now returns a Retriever that supports `traversal` as a search type.

libs/knowledge-store/ragstack_knowledge_store/document_graph.py
cbornet: Removed links from the API for the moment as explicit links can be defined in metadata (see ExplicitEdgeExtractor)

libs/knowledge-store/ragstack_knowledge_store/cassandra.py
```diff
@@ -245,28 +248,20 @@ def embeddings(self) -> Optional[Embeddings]:
...
         return ConcurrentQueries(self._session, concurrency=self._concurrency)
 
-    # TODO: async
-    def add_texts(
+    def add_nodes(
         self,
-        texts: Iterable[str],
-        metadatas: Optional[Iterable[Dict[str, Any]]] = None,
+        nodes: Iterable[Node] = None,
+        links: Iterable[Tuple[str, str]] = None,
```
bjchambers, line 254: Links is no longer part of the signature.

libs/knowledge-store/ragstack_knowledge_store/cassandra.py
```diff
@@ -362,65 +366,62 @@ def add_href_url_pairs(href_ids, url_ids):
...
-        ids = kwargs.pop("ids")
-        knowledge_store = KnowledgeStore(embedding, **kwargs)
-        knowledge_store.add_texts(texts, metadatas, ids=ids)
-        return knowledge_store
+    @classmethod
+    def from_documents(
+        cls: Type["CassandraKnowledgeStore"],
+        documents: Iterable[Document],
+        embedding: Embeddings,
+        ids: Optional[Iterable[str]] = None,
```
bjchambers, line 388: Feels like this should take `metadatas` as well?

libs/knowledge-store/ragstack_knowledge_store/cassandra.py
```diff
@@ -447,8 +448,13 @@ def _linked_ids(
         results = self._session.execute(self._query_linked_ids, (source_id,))
         return _results_to_ids(results)
 
-    def retrieve(
-        self, query: Union[str, Iterable[str]], *, k: int = 4, depth: int = 1
+    def traversing_retrieve(
+        self,
+        query: Union[str, Iterable[str]],
```
bjchambers, line 453: I think for simplicity / consistency we could probably have this take `query: str`. Thinking about it, it's going to be difficult for more sophisticated retrievers to support the union directly, and that should probably be implemented as part of the things that will actually need to take multiple queries.

bjchambers (approved):

libs/knowledge-store/ragstack_knowledge_store/base.py
```diff
@@ -0,0 +1,345 @@
...
+        node_id,
+    ) in zip(documents, ids):
+        yield TextNode(
+            id=node_id,
+            metadata=doc.metadata,
+            text=doc.page_content,
+        )
+
+
+class KnowledgeStore(VectorStore):
```
bjchambers, line 70: I think we'll want to iterate on naming (`KnowledgeStore`, `ContentStore`, `DocumentStore`, etc.) have all come up. But that should be in a separate PR (don't want to block the functionality improvements in other PRs which are going to be used in upcoming posts, etc.)

libs/knowledge-store/ragstack_knowledge_store/base.py
```diff
@@ -0,0 +1,345 @@
...
+        self,
+        texts: Iterable[str],
+        metadatas: Optional[Iterable[dict]] = None,
+        *,
+        ids: Optional[Iterable[str]] = None,
+        **kwargs: Any,
+    ) -> List[str]:
+        metadatas_it = iter(metadatas) if metadatas else itertools.repeat({})
+        ids_it = iter(ids) if ids else itertools.repeat(None)
+        nodes = _texts_to_nodes(iter(texts), metadatas_it, ids_it)
```
bjchambers, line 124: May want `_texts_to_nodes` to take the iterables and do the repeat. Specifically, there is the following issue:

1. We'd like to ensure that if presented, the metadatas, ids and texts all have the same lengths. This suggests `zip(texts, metadatas, ids, strict=True)`.
2. That will fail when presented the infinite repeat iterators.
3. So `_texts_to_nodes` will need to look at which of them are present to decide which to zip over.

libs/knowledge-store/ragstack_knowledge_store/cassandra.py
cbornet: Metadatas is a field of `Document`

libs/knowledge-store/ragstack_knowledge_store/cassandra.py
cbornet: done

libs/knowledge-store/ragstack_knowledge_store/cassandra.py
cbornet: done

libs/knowledge-store/ragstack_knowledge_store/base.py
cbornet: `strict` was added to Python 3.10 so I think we should refrain from using it atm.
LangChain is compatible with Python 3.8.1+

cbornet: Done (removing zip and checking iterators manually)

PR #459, opened 2024-06-03
[Docs] - Knowledge graph

mendonk:
* Index doc for overview of graph-based ragstack libraries
* Doc for knowledge graph library
* Doc for local run of graph notebook
* Links to notebook at /datastax-labs (should we use the one in the package library instead?)
* Nav
* Example of graph rag with yaml schema
Knowledge store in progress.


docs/modules/examples/pages/knowledge-graph.adoc
```diff
@@ -0,0 +1,288 @@
+= Knowledge Graph
+
+image::https://colab.research.google.com/assets/colab-badge.svg[align="left",link="https://colab.research.google.com/github/datastax-labs/knowledge-graphs-langchain/blob/main/notebook.ipynb"]
+
+Use RAGStack, https://python.langchain.com/docs/use_cases/graph/constructing/#llm-graph-transformer[`LLMGraphTransformer`], and https://www.datastax.com/products/datastax-astra[DataStax AstraDB] to extract knowledge triples and store them in a vector database.
```
bjchambers, line 5: It's worth noting there is an alternative way to extract that doesn't use their `LLMGraphTransformer`, and instead uses a `KnowledgeSchema` (something we introduced) that let's you describe the kind of nodes and edges to be extracted.

This test demonstrates it's functionality:

https://github.com/datastax/ragstack-ai/blob/main/libs/knowledge-graph/tests/test_extraction.py

You can define the extractor in code, or via a YAML file:

https://github.com/datastax/ragstack-ai/blob/main/libs/knowledge-graph/tests/marie_curie_schema.yaml

docs/modules/examples/pages/knowledge-graph.adoc
```diff
@@ -0,0 +1,288 @@
...
+----
+pip install ragstack-ai-knowledge-graph python-dotenv
+----
++
+. Create a `.env` file and store the necessary credentials.
++
+[source,bash]
+----
+OPENAI_API_KEY="sk-..."
+ASTRA_DB_DATABASE_ID="670d40c2-80f9-4cb0-8c74-d524dd6944d1"
```
bjchambers, line 27: Do we really want a database ID showing up? Is this a common pattern?

docs/modules/examples/pages/knowledge-graph.adoc
```diff
@@ -0,0 +1,288 @@
...
++
+. Render the extracted graph to GraphViz and save the extracted graph documents to the AstraDB graph store.
++
+[source,python]
+----
+render_graph_documents(graph_documents)
+graph_store.add_graph_documents(graph_documents)
+----
+
+=== Optional: Predefine relationships between entities
```
bjchambers, line 102: Can probably drop this -- the knowledge schema stuff is more robust, and we can add a notebook / docs on that at some point.

docs/modules/knowledge-graph/pages/index.adoc
```diff
@@ -0,0 +1,44 @@
+= Graph-based Knowledge Extraction and Traversal
+
+RAGStack offers two libraries supporting knowledge graph extraction and traversal, `ragstack-ai-knowledge-graph` and `ragstack-ai-knowledge-store`.
+
+A knowledge graph represents distinct entities (real world objects, like "Marie Curie" or "maize") as **nodes**. Nodes are connected by **edges** indicating relationships between them. Each edge includes the source (for example, "Marie Curie" the person), the target ("Nobel Prize" the award) and a type, indicating how the source relates to the target (for example, “won”).
```
bjchambers, line 5: coarse grained (knowledge store) the nodes are less entities and more specific content / passages / etc.

not sure we need to reflect both completely, but JFYI

docs/modules/knowledge-graph/pages/index.adoc
```diff
@@ -0,0 +1,44 @@
...
+
+A graph database isn't required to use the knowledge graph libraries - RAGStack uses Astra DB or Apache Cassandra to store and retrieve graphs.
+
+*TL;DR:*
```
bjchambers, line 9: The TL;DR seems longer than the original? Feels like this is less of a summary and rather more information?

docs/modules/knowledge-graph/pages/index.adoc
```diff
@@ -0,0 +1,44 @@
...
+
+. Retrieval augmented generation based on chunking and vector similarity search has some weaknesses.
+.. Similarity search looks for information most similar to the question. This makes is harder to answer questions with multiple topics.
```
bjchambers, line 12: Another way of explaining this that I've used:

1. Similarity search only finds similar information.
2. There are many cases where less-similar information is relevant to the question. Consider a tech support system where you reach an article which *is* similar, that says "if you have trouble with step 4, see <blah>". Even if `<blah>` is *not* similar to the question, it likely provides more information -- the author even says "see also". This is an easy edge to understand (from the coarse grained graph case)
3. Alternatively, if you retrieve 100 chunks that are all highly similar to the question, you have likely retrieved 100 chunks that are also highly similar to themselves, even if you employ MMR. Following edges to "linked" information can allow you to increase the diversity of the results, which provides more distinct information (when combined).

docs/modules/knowledge-graph/pages/index.adoc
```diff
@@ -0,0 +1,44 @@
...
+[source,bash]
+====
+pip install ragstack-ai-knowledge-graph
+====
+
+For more information, see xref:knowledge-graph.adoc[].
+
+== The `ragstack-ai-knowledge-store` library
+
+The `ragstack-ai-knowledge-store` library contains functions for creating a hybrid vector-and-graph knowledge store. This store combines the benefits of vector search with the context and relationships of a knowledge graph by describing the relationships between Document chunks with edges.
```
bjchambers, line 32: Alternative ways to present knowledge store:

1. It's a knowledge graph with a general purpose schema represent the content of your unstructured documents. Instead of identifying specific entities, the nodes correspond to things like "text passages" (chunks) and edges capture relationships between those passages such as "links to". This is much more scalable to construct, because you don't need a specific schema or an LLM to extract information, and it is easier to consume by an LLM.
2. It's a vector store to which we've added "related edges" or "see also" paths. This brings the benefits of a knowledge graph (providing relevant, non-semantic similar information) without requiring the extra work of extracting a detailed knowledge graph. It can (conceptually/theoretically/not yet implementedly) be "added" to a vector implementation.

docs/modules/knowledge-graph/pages/knowledge-graph.adoc
```diff
@@ -0,0 +1,101 @@
...
+The `ragstack-ai-knowledge-graph` library contains functions for the extraction and traversal of knowledge graphs.
+
+To install the library, run:
+
+[source,bash]
+====
+pip install ragstack-ai-knowledge-graph
+====
+
+== How is Knowledge Graph different from RAG?
```
bjchambers, line 18: Maybe worth starting out with: It isn't. It is a different representation for storing information, but can still be used to retrieve and augment generated results.

Specifically, vector is a way to do RAG, but knowledge graphs are too.

bjchambers (approved):

docs/modules/examples/pages/knowledge-graph.adoc
mendonk: It's fake

NadirJ (approved):

docs/modules/examples/pages/knowledge-graph.adoc
mendonk: Added

docs/modules/knowledge-graph/pages/knowledge-graph.adoc
mendonk: Added, thanks for clarification

PR #460, opened 2024-06-03
[knowledge-store] feat: Dynamic, MMR-based traversal

bjchambers:
This required denormalizing the `text_embedding` of target nodes into
the edge, making it easier to guide traversal based on distance to a
query.
    
Also fixed a bug in keyword linking which didn't create links to old
nodes with the given keyword.


libs/knowledge-store/ragstack_knowledge_store/knowledge_store.py
```diff
@@ -148,6 +183,14 @@ def __init__(
             """
         )
 
+        self._query_linked_ids_and_embedding = session.prepare(
+            f"""
+            SELECT target_content_id, target_embedding AS content_id
```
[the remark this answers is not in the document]
kerinin: This seems OK to me - we're already optimizing for query efficiency over insertion efficiency.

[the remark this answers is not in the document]
bjchambers: Agreed. Also, many of the edge extractors will do something like "query for nodes with this keyword", which can be written to written to return the embedding, so in most cases it won't add any extra steps.

libs/knowledge-store/ragstack_knowledge_store/knowledge_store.py
```diff
@@ -46,6 +49,39 @@ def _results_to_ids(results: Optional[ResponseFuture]) -> Iterable[str]:
...
+  embedding: np.ndarray
+  """Embedding used for updating similarity to selections."""
+
+  redundancy: float
+  """(1 - Lambda) * max(Similarity to selected items)."""
+
+  def __init__(self, embedding: List[float], lambda_mult: float, query_embedding: np.ndarray):
+    self.embedding = emb_to_ndarray(embedding)
+
+    # TODO: Refactor to use cosine_similarity_top_k to allow an array of embeddings?
```
bjchambers, line 73: I need to rework somethings to make this take advantage of the fact that cosine_similarity operates on a matrix (I believe) so that I can do all of the comparisons in one step. But, this works for now. Should that be deferred, or is it worth doing in this iteration?

bjchambers:
@kerinin @cbornet this should be ready for review. I left two comments about things that may be worth doing in this iteration (vectorized cosine-similarity computation and score thresholds). Otherwise, it is tested and working. Things that may be especially worth looking at:

- Naming / logic in the loops (extractors and mmr traversal). This was somewhat tricky, and I'm sure we can improve readability. I'll revisit when I have time too.
- Changes to edge extractors to denormalize the embeddings into the edges.

libs/knowledge-store/ragstack_knowledge_store/knowledge_store.py
cbornet: How about using `langchain_community.utils.math.cosine_similarity` ? It uses numpy and has optional support for simd.

libs/knowledge-store/ragstack_knowledge_store/knowledge_store.py
```diff
@@ -342,28 +417,127 @@ def _linked_ids(
...
+                Defaults to 10.
+            max_depth: Maximum depth of a node (number of edges) from a node retrieved
+                via similarity. Defaults to 2.
+            lambda_mult: Number between 0 and 1 that determines the degree
+                of diversity among the results with 0 corresponding to maximum
+                diversity and 1 to minimum diversity.
+                Defaults to 0.5.
+        """
+
+        # TODO: Add score_threshold?
```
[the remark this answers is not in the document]
cbornet: There is a `score_threshold` param in the VectorStore `similarity_search_with_relevance_scores`

libs/knowledge-store/ragstack_knowledge_store/knowledge_store.py
cbornet: Oh, right. I read too quickly.

libs/knowledge-store/ragstack_knowledge_store/knowledge_store.py
```diff
@@ -342,28 +420,127 @@ def _linked_ids(
         results = self._session.execute(self._query_linked_ids, (source_id,))
         return _results_to_ids(results)
 
+    def mmr_traversal(
```
cbornet, line 423: atm, we call the method `traversing_retrieve` in #452 .
Should we call it `traversal` for consistency ?

bjchambers: I think we should figure out a pattern, but it can wait until we start stabilizing things. I think maybe they should all have retrieve, but in VectorStore they have `similarity_search` which doesn't? So maybe these are instead `traverse` and `max_marginal_relevance_traverse`?

Looking at https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/vectorstores.py#L164, we should also make them available via `search`, and that shows the different names of the specific methods...

bjchambers: Right, but the question is still the method names (and to some degree how that maps to names in `search`).

I think that `similarity_search` as a method name is meant as a verb (you do "similarity search") rather than a noun (you perform "a similarity search"). Since they also have `max_marginal_relevance_search`, it makes me wonder:

I think either of these make sense:

* `traverse` and `max_marginal_relevance_traverse` -- the thing we do is "traverse" rather than "search"
* `traversing_search` and `max_marginal_relevance_traversing_search` -- the thing we do is still search, but we add more modifiers. these names feel long...
* maybe `traversal_search` and `max_marginal_relevance_traversal_search` -- using traversal instead of traversing

For `search_type` it seems like they omit the `_search` from the method name (and abbreviate things). So that would suggest:

* method: `traversing_search`, `search_type: 'traversing'`
* method: `mmr_traversing_search` (or `max_marginal_relevance_traversing_search`), `search_type: `mmr_traversing`)

(although `traversal` also seems good as the word).

I don't have strong opinions beyond consistency (probably `_search` suffix makes sense, etc.) -- any thoughts on your side?

cbornet: I like traversal_search

libs/knowledge-store/ragstack_knowledge_store/knowledge_store.py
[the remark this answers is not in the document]
bjchambers: added `score_threshold`. not sure it's needed yet but wasn't too hard.

I could try to combine 1xN cosine similarity, but that would require changing the loop. I think I'll leave that part as a todo.

libs/knowledge-store/ragstack_knowledge_store/base.py
```diff
@@ -210,19 +210,96 @@ async def atraversing_retrieve(
...
+        ):
+            yield doc
+
+    @abstractmethod
+    def mmr_traversal_search(
+        self,
+        query: str,
+        *,
+        k: int = 4,
+        max_depth: int = 2,
```
cbornet, line 223: Why is it different than in `traversal_search` ? Can we use `depth` here too ?

libs/knowledge-store/ragstack_knowledge_store/cassandra.py
```diff
@@ -58,8 +94,10 @@ def __init__(
         apply_schema: bool = True,
         concurrency: int = 20,
     ):
-        """
-        Create the hybrid knowledge store.
+        """A hybrid vector-and-graph knowledge store backed by Cassandra.
+
+        Document chunks support vector-similarity search as well as edges linking
+        documetns based on structural and semantic properties.
```
cbornet, line 100:
```suggestion
        documents based on structural and semantic properties.
```

libs/knowledge-store/ragstack_knowledge_store/directed_edge_extractor.py
```diff
@@ -60,73 +63,66 @@ def _targets(self, metadata: Dict[str, Any]) -> Set[str]:
...
     ) -> int:
         # First, iterate over the new nodes, collecting the sources/targets that
         # are referenced and which IDs contain those.
         new_ids = set()
-        new_sources_to_ids = {}
-        new_targets_to_ids = {}
-        for md in metadatas:
+        resource_to_new_defs_embs = {}
+        resource_to_new_refs = {}
+        for (md, embedding) in zip(metadatas, text_embeddings, strict=True):
```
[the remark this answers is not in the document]
cbornet: An equivalent can be done with 
```python
sentinel = object()
itertools.zip_longest(metadatas, text_embeddings, fillvalue=sentinel)
```
then raising if one of the values is `sentinel`

libs/knowledge-store/ragstack_knowledge_store/base.py
bjchambers: We could, but they are different. In `traversal_search` it pulls back *everything* up to the specified `depth`. In `mmr_traversal_search` this acts as a limit on the maximum depth of anything it can pull back, but it doesn't "eagerly" pull things back to that depth.

So, the thinking is that in the `traversal_search` case it specifies the depth to act on, while here it specifies a limit (maximum) on the depth.

cbornet: OK. It's just that I can imagine users going back and forth between traversal and mmr-traversal and not noticing that the param name is not the same.

NadirJ (approved):

libs/knowledge-store/ragstack_knowledge_store/base.py
bjchambers: That makes sense. I can switch it. My concern was "same name, different behavior", but I think that can be fine here.

PR #461, opened 2024-06-06
[knowledge store] docs: Astra Docs Demo

bjchambers:
This creates a demo notebook that loads the Astra
docs into a KnowledgeStore with edges inferred
from links in the content.

This revealed a bug in the directed edge extractor which is fixed.

### DIFF
--- a/libs/knowledge-store/notebooks/astra_support.ipynb
+++ b/libs/knowledge-store/notebooks/astra_support.ipynb
@@ -582,24 +582,29 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 30,
+   "execution_count": 33,
    "metadata": {},
    "outputs": [
     {
      "data": {
       "text/markdown": [
-       "Astra DB Serverless uses multiple indexing techniques to facilitate efficient vector search, including the JVector vector search engine and Storage-Attached Index (SAI).\n",
+       "Astra DB Serverless uses multiple indexing techniques to speed up vector searches, with a focus on efficient and scalable solutions. Specifically, Astra employs the following vector indexing algorithms:\n",
        "\n",
-       "1. **JVector:**\n",
-       "   - **Type:** Graph-based indexing\n",
-       "   - **Family:** DiskANN\n",
-       "   - **Functionality:** JVector constructs a graph index to enable efficient vector search. It supports incremental updates, meaning new documents can be added to the graph immediately, allowing for efficient searches right away. JVector also allows for compression of vectors using quantization to save space and improve performance.\n",
+       "1. **JVector**:\n",
+       "   - **Graph Index**: Astra DB uses the [JVector vector search engine](https://github.com/jbellis/jvector) to construct a graph index. JVector is a graph-based index that builds on the DiskANN design with composable extensions.\n",
+       "   - **Graph-based Indexes**: These indexes are simpler to implement and typically faster. They can be constructed and updated incrementally, making them more suitable for general-purpose indexing compared to partition-based approaches.\n",
+       "   - **On-Disk Adjacency List**: The graph is represented by an on-disk adjacency list per node, with additional data stored inline to support two-pass searches.\n",
+       "   - **Two-Pass Searches**: The first pass uses lossily compressed representations of the vectors kept in memory, and the second pass uses a more accurate representation read from disk.\n",
        "\n",
-       "2. **Storage-Attached Index (SAI):**\n",
-       "   - **Type:** General indexing technique for various data types, including vectors\n",
-       "   - **Functionality:** SAI indexes are attached to the SSTables (Sorted Strings Tables) that store the rows of data, allowing for efficient search of non-partition columns. SAI provides numeric-, text-, and vector-based indexes to support different kinds of searches. It can be customized based on requirements, such as a specific similarity function or text transformation.\n",
+       "2. **Product Quantization (PQ)**:\n",
+       "   - **Compression**: JVector can compress vectors using product quantization to save space and improve performance.\n",
+       "   - **Anisotropic Weighting**: Optionally used with product quantization to improve recall.\n",
        "\n",
-       "These indexing methods ensure fast and efficient similarity searches in high-dimensional vector spaces, making Astra DB Serverless suitable for applications requiring vector search capabilities."
+       "3. **Storage-Attached Index (SAI)**:\n",
+       "   - **Efficient Searching**: SAI is an indexing technique used to efficiently find rows that satisfy query predicates. It supports numeric-, text-, and vector-based indexes.\n",
+       "   - **Customization**: You can customize indexes based on specific requirements, such as using a specific similarity function or text transformation.\n",
+       "\n",
+       "These indexing techniques enable Astra DB Serverless to provide efficient and scalable vector search capabilities, suitable for various high-dimensional data scenarios often encountered in machine learning and AI applications."
       ],
       "text/plain": [
        "<IPython.core.display.Markdown object>"
--- a/libs/knowledge-store/notebooks/astra_support.ipynb
+++ b/libs/knowledge-store/notebooks/astra_support.ipynb
@@ -1,6 +1,16 @@
 {
  "cells": [
   {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%load_ext autoreload\n",
+    "%autoreload 2"
+   ]
+  },
+  {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
@@ -9,7 +19,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": 1,
    "metadata": {},
    "outputs": [
     {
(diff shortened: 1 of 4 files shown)


libs/knowledge-store/ragstack_knowledge_store/directed_edge_extractor.py
```diff
@@ -101,7 +101,7 @@ def add_source_target_pairs(href_ids, url_ids):
                     for url_id in url_ids:
                         if not isinstance(url_id, str):
                             url_id = url_id.content_id
-                    source_target_pairs.add((href_id, url_id))
+                        source_target_pairs.add((href_id, url_id))
```
bjchambers, line 104: I may try to look at the tests to see if I can handle this case -- my guess is it wasn't getting hit for some reason.

libs/knowledge-store/notebooks/support.ipynb
```diff
@@ -0,0 +1,596 @@
...
+   "outputs": [],
+   "source": [
+    "from langchain_community.document_loaders import AsyncHtmlLoader\n",
+    "from bs4 import BeautifulSoup\n",
+    "from langchain_core.documents import Document\n",
+    "from typing import AsyncIterator, Iterable, Set\n",
+    "from markdownify import MarkdownConverter\n",
+    "from urllib.parse import urlparse, urljoin, urldefrag\n",
+    "\n",
+    "def parse_url(link,\n",
```
bjchambers, line 137: Would be good to move some of these larger code blocks into the library. At least things like "get all the `a href` links in an html body" should be pretty re-usable. I'll probably do that in a follow-up.

libs/knowledge-store/notebooks/support.ipynb
```diff
@@ -0,0 +1,596 @@
...
+     "text": [
+      "Vector Only: https://docs.datastax.com/en/cql/astra/getting-started/vector-search-quickstart.html\n",
+      "Vector Only: https://docs.datastax.com/en/astra-db-serverless/get-started/concepts.html\n"
+     ]
+    },
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Traversing:  https://docs.datastax.com/en/cql/astra/getting-started/vector-search-quickstart.html\n",
```
bjchambers, line 454: Note that traversing gives us a lot of output. I suspect we would benefit from the MMR retrieval, and would like to add that to this notebook once it goes in.

libs/knowledge-store/notebooks/support.ipynb
```diff
@@ -0,0 +1,596 @@
...
+      "Requirement already satisfied: soupsieve>1.2 in /Users/benjamin.chambers/Library/Caches/pypoetry/virtualenvs/agent-framework-aiP65pJh-py3.11/lib/python3.11/site-packages (from beautifulsoup4<5,>=4.9->markdownify) (2.5)\n",
+      "\n",
+      "\u001b[1m[\u001b[0m\u001b[34;49mnotice\u001b[0m\u001b[1;39;49m]\u001b[0m\u001b[39;49m A new release of pip is available: \u001b[0m\u001b[31;49m23.3.1\u001b[0m\u001b[39;49m -> \u001b[0m\u001b[32;49m24.0\u001b[0m\n",
+      "\u001b[1m[\u001b[0m\u001b[34;49mnotice\u001b[0m\u001b[1;39;49m]\u001b[0m\u001b[39;49m To update, run: \u001b[0m\u001b[32;49mpip install --upgrade pip\u001b[0m\n",
+      "Note: you may need to restart the kernel to use updated packages.\n"
+     ]
+    }
+   ],
+   "source": [
+    "%pip install markdownify python-dotenv"
```
cbornet, line 32: Should we add `ragstack-ai-knowledge-store` ?

libs/knowledge-store/notebooks/support.ipynb
```diff
@@ -0,0 +1,596 @@
...
+    "session.execute(f\"DROP TABLE IF EXISTS {keyspace}.{SITE_PREFIX}_nodes\")\n",
+    "session.execute(f\"DROP TABLE IF EXISTS {keyspace}.{SITE_PREFIX}_edges\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Load the Documents\n",
+    "Finally, we fetch pages and write them to the knowledge store in batches of 50."
```
cbornet, line 320: Shouldn't the store handle the batching ?

libs/knowledge-store/notebooks/support.ipynb
```diff
@@ -0,0 +1,596 @@
...
+      "Traversing:  https://docs.datastax.com/en/astra-db-serverless/integrations/embedding-providers/nvidia-nemo.html\n",
+      "Traversing:  https://docs.datastax.com/en/astra-db-serverless/integrations/embedding-providers/openai.html\n",
+      "Traversing:  https://docs.datastax.com/en/astra-db-serverless/integrations/embedding-providers/azure-openai.html\n",
+      "Traversing:  https://docs.datastax.com/en/astra-db-serverless/get-started/examples.html\n"
+     ]
+    }
+   ],
+   "source": [
+    "# Set the question and see what documents each technique retrieves.\n",
+    "QUESTION=\"What vector indexing algorithms does Astra use?\"\n",
```
cbornet, line 490: Could we ask a question that is not answered correctly with `depth=0` and is correctly answered with a higher depth ?

libs/knowledge-store/notebooks/support.ipynb
```diff
@@ -0,0 +1,596 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 7,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "Fetching pages:  22%|##1       | 300/1368 [00:51<06:37,  2.68it/s]Error fetching https://docs.datastax.com/en/cql/dse/reference/cql-commands/alter-materialized-view.html with attempt 1/3: Cannot connect to host docs.datastax.com:443 ssl:default [nodename nor servname provided, or not known]. Retrying...\n",
```
cbornet, line 332: Can we have a run without errors ?

libs/knowledge-store/notebooks/support.ipynb
```diff
@@ -0,0 +1,596 @@
...
+    "    import dotenv\n",
+    "    dotenv.load_dotenv()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Initialize Cassio and Knowledge Store\n",
+    "With the environent variables set we initialize the Cassio library for talking to Cassandra / Astra DB.\n",
```
cbornet, line 264:
```suggestion
    "With the environment variables set we initialize the Cassio library for talking to Cassandra / Astra DB.\n",
```

libs/knowledge-store/notebooks/support.ipynb
bjchambers: (right now it gets it from the environment -- this is part of what I'll need to test to make it work in colab, but I also want to start using MMR which means that PR needs to go in and be released before it will be available via pip install)

My plan is to merge MMR, merge this, make sure MMR works locally, then do a release, then create a PR updating the notebook.

libs/knowledge-store/notebooks/support.ipynb
bjchambers: Maybe? It currently doesn't (it currently tries to traverse the whole iterable). Further, it would need to take an async iterable in the async case, which I'm not sure it does. Plan is to get this in, and then we can look at the notebook and see if we have ways to simplify it (better to have something that works / demonstrates functionality and then simplify by improving the code).

libs/knowledge-store/notebooks/support.ipynb
bjchambers: This question is not answered correctly with depth 0 and is correctly answered with higher depth, which is why this is the question that is used.

libs/knowledge-store/notebooks/support.ipynb
bjchambers: Maybe? I don't want to keep rerunning this since it is a lot of embeddings / writes to Astra. I'd like to get this in as a baseline and then work on a final version once I add MMR, etc.

libs/knowledge-store/notebooks/support.ipynb
bjchambers: https://github.com/datastax/ragstack-ai/blob/main/libs/knowledge-store/ragstack_knowledge_store/cassandra.py#L215

Specifically, the above will consume the iterable and try to put all the content in memory simultaneously. Agreed we should probably improve that (or at least document it). But even if it did, it wouldn't work with an `AsyncIterator`, etc. So, there is probably a case to be made that batching (and interacting with concurrency in useful ways) needs to happen outside of the client.

libs/knowledge-store/notebooks/support.ipynb
cbornet: Maybe you could emphasize that with an analysis of the results ?

NadirJ (approved):

PR #462, opened 2024-06-07
Support Python 3.9 in knowledge_store and knowledge_graph

cbornet:


libs/knowledge-store/ragstack_knowledge_store/cassandra.py
```diff
@@ -222,7 +222,7 @@ def add_nodes(
 
         ids = []
         with self._concurrent_queries() as cq:
-            tuples = zip(texts, text_embeddings, metadatas, strict=True)
+            tuples = zip(texts, text_embeddings, metadatas)
```
cbornet, line 225: We don't need to check the iterators lengths as they are equal by construction.

bjchambers (approved):

nicoloboschi (approved):

PR #465, opened 2024-06-07
Use setup_mode instead of apply_schema for consistency with other components

cbornet:


libs/knowledge-store/ragstack_knowledge_store/cassandra.py
```diff
@@ -85,8 +86,12 @@ def __init__(
         self._session = session
         self._keyspace = keyspace
 
-        if apply_schema:
+        if setup_mode == SetupMode.SYNC:
             self._apply_schema()
+        elif setup_mode != SetupMode.OFF:
+            raise ValueError(
+                "Invalid setup mode. Only SYNC and OFF are supported at the moment"
```
bjchambers, line 93: Maybe display the mode tha was chosen so we know?

bjchambers (approved):

libs/knowledge-store/ragstack_knowledge_store/cassandra.py
cbornet: done

PR #466, opened 2024-06-07
llamaindex: pin llama-index-core version

nicoloboschi:


libs/llamaindex/pyproject.toml
```diff
@@ -18,6 +18,7 @@ ragstack-ai-colbert = "1.0.5"
 
 # llama-index
 llama-index = "0.10.31"
+llama-index-core = "0.10.43"
```
epinzur, line 21: maybe match 0.10.31 of llama-index

epinzur (approved):

PR #467, opened 2024-06-07
Extract LangChain knowledge store to its own module

cbornet:
First step.
Next step will be to put it in ragstack-langchain package.


libs/knowledge-store/ragstack_knowledge_store/knowledge_store.py
```diff
@@ -1,47 +1,65 @@
...
 
 import numpy as np
 from cassandra.cluster import ConsistencyLevel, ResponseFuture, Session
 from cassio.config import check_resolve_keyspace, check_resolve_session
-from langchain_community.utilities.cassandra import SetupMode
-from langchain_community.utils.math import cosine_similarity
-from langchain_core.documents import Document
-from langchain_core.embeddings import Embeddings
 
 from ragstack_knowledge_store.edge_extractor import EdgeExtractor
```
bjchambers, line 16: Why do we mix `ragstack_knowledge_store.edge_extractor` and `._utils`? Should we use all relative or all absolute imports?

libs/knowledge-store/ragstack_knowledge_store/langchain/cassandra.py
```diff
@@ -0,0 +1,198 @@
...
+            edge_extractors=edge_extractors,
+            node_table=node_table,
+            edge_table=edge_table,
+            session=self._session,
+            keyspace=_keyspace,
+            setup_mode=_setup_mode,
+            concurrency=concurrency,
+        )
+
+        self._query_by_embedding = self._session.prepare(
```
bjchambers, line 98: I feel like the query by embedding could/should be in the `knowledge_store.py`. We could have a "get nearest nodes", and then have LangChain wrap that.

libs/knowledge-store/ragstack_knowledge_store/langchain/cassandra.py
```diff
@@ -0,0 +1,198 @@
...
+        with the same `source` metadata value.
+        Args:
+            embedding: The embeddings to use for the document content.
+            edge_extractors: Edge extractors to use for linking knowledge chunks.
+            concurrency: Maximum number of queries to have concurrently executing.
+            setup_mode: Mode used to create the Cassandra table (SYNC,
+                ASYNC or OFF).
+        """
+        self._embedding = embedding
+        self._session = check_resolve_session(session)
```
bjchambers, line 83: I don't think this file (`cassandra.py`) should have the session/embedding/etc. That should all be part of `self.store` only.

libs/knowledge-store/ragstack_knowledge_store/langchain/cassandra.py
```diff
@@ -0,0 +1,198 @@
...
+    def embeddings(self) -> Optional[Embeddings]:
+        return self._embedding
+
+    def add_nodes(
+        self,
+        nodes: Iterable[Node] = None,
+        **kwargs: Any,
+    ):
+        _nodes = []
+        for node in nodes:
```
bjchambers, line 117: Why does this need code? It would probably make more sense to just call `self.store.add_nodes(nodes)` and let that handle the kinds of nodes that are supported or not.

libs/knowledge-store/ragstack_knowledge_store/langchain/cassandra.py
```diff
@@ -0,0 +1,198 @@
...
+        embedding_vector = self._embedding.embed_query(query)
+        return self.similarity_search_by_vector(
+            embedding_vector,
+            k=k,
+        )
+
+    def similarity_search_by_vector(
+        self, embedding: List[float], k: int = 4, **kwargs: Any
+    ) -> List[Document]:
+        results = self._session.execute(self._query_by_embedding, (embedding, k))
```
bjchambers, line 166: This should probably be:

```
return [_node_to_document(node)
        for node in self.store.similarity_search_by_vector(...)]`
```

libs/knowledge-store/ragstack_knowledge_store/langchain/cassandra.py
cbornet: This code makes the adaptation from LangChain's `Node` to RagStack's `Node`.
It is required if we don't want one layer to leak into the other.

libs/knowledge-store/ragstack_knowledge_store/langchain/cassandra.py
cbornet: Sure.

libs/__init__.py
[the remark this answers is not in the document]
cbornet: I have no idea how this file got there 😅

libs/knowledge-store/ragstack_knowledge_store/langchain/cassandra.py
```diff
@@ -0,0 +1,198 @@
...
+    )
+
+
+def _results_to_documents(results: Optional[ResponseFuture]) -> Iterable[Document]:
+    if results:
+        for row in results:
+            yield _row_to_document(row)
+
+
+class CassandraKnowledgeStore(KnowledgeStore):
```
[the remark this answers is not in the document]
cbornet: done.

libs/__init__.py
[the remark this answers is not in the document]
cbornet: fixed.

libs/knowledge-store/ragstack_knowledge_store/knowledge_store.py
cbornet: fixed.

libs/knowledge-store/ragstack_knowledge_store/langchain/cassandra.py
cbornet: done.

libs/knowledge-store/ragstack_knowledge_store/langchain/cassandra.py
cbornet: done.

libs/knowledge-store/ragstack_knowledge_store/langchain/cassandra.py
cbornet: done.

libs/knowledge-store/ragstack_knowledge_store/knowledge_store.py
```diff
@@ -1,49 +1,67 @@
...
 
-def _row_to_document(row) -> Document:
-    return Document(
-        page_content=row.text_content,
+@dataclass
+class Node:
+    """Node in the KnowledgeStore graph"""
+
+    id: Optional[str] = None
+    """Unique ID for the node. Shall be generated by the KnowledgeStore if not set"""
```
bjchambers, line 33: I think `Will be generated` -- we're telling the user what will happen, rather than imposing a requirement on the implementation.

libs/knowledge-store/ragstack_knowledge_store/knowledge_store.py
```diff
@@ -1,49 +1,67 @@
...
+    metadata: dict = field(default_factory=dict)
+    """Metadata for the node. May contain information used to link this node 
+    with other nodes."""
+
+
+@dataclass
+class TextNode(Node):
+    text: str = None
+    """Text contained by the node"""
```
bjchambers, line 42: Does PyDoc require ending in a `.` like Javadoc?

libs/knowledge-store/ragstack_knowledge_store/knowledge_store.py
```diff
@@ -1,49 +1,67 @@
...
+
+
+class SetupMode(Enum):
```
bjchambers, line 45: I wonder if this is too much LangChain-ism in the base. What if we just had methods for setup (`setup_schema(...)`, `asetup_schema(...)`, etc.). Then the langchain code can choose which to call based on the setup mode from LangChain, but we don't need to mirror the enum into base?

libs/knowledge-store/ragstack_knowledge_store/langchain/cassandra.py
```diff
@@ -0,0 +1,194 @@
...
+from langchain_core.documents import Document
+from langchain_core.embeddings import Embeddings
+
+from .base import KnowledgeStore, Node, TextNode
+from ragstack_knowledge_store import knowledge_store
+from ragstack_knowledge_store.edge_extractor import EdgeExtractor
+from ragstack_knowledge_store.embedding_model import EmbeddingModel
+
+
+class EmbeddingModelAdapter(EmbeddingModel):
```
bjchambers, line 20: Should this be `_EmbeddingModelAdapter`? IIUC this isn't meant for public consumption.

libs/knowledge-store/ragstack_knowledge_store/langchain/cassandra.py
bjchambers: Ah. That makes sense. I guess the other option would be for us to hide the non-text nodes from LangChain (eg., not have a `Node` or `add_nodes` yet). That may reduce the apparent size of the interface -- we could still say the underlying thing has support for multi-modal nodes, and come back and add that once it's in LangChain.

Either way seems OK though if we think the additional node isn't going to add complexity.

libs/knowledge-store/ragstack_knowledge_store/math.py
```diff
@@ -0,0 +1,42 @@
+"""Copied from langchain_community.utils.math"""
```
bjchambers, line 1: maybe worth mentioning the commit and/or adding a link (thinking about hte future where we may want to see if there are bug fixes, etc.)

bjchambers (approved):

libs/knowledge-store/ragstack_knowledge_store/knowledge_store.py
cbornet: We could. But our strategy until now in cassio/astrapy/etc... has been to default to creating the schema in the constructor to minimize the number of operations to do in notebooks.
We can change that here.

libs/knowledge-store/ragstack_knowledge_store/math.py
cbornet: Done. Even better would be to write our own code. Even if this is MIT licensed.

PR #471, opened 2024-06-10
feat: Use tags exclusively for edge creation

bjchambers:
This enables edge extractors to be implemented as
document transformers configuring the metadata,
rather than needing to interact with the
underlying storage.


libs/knowledge-store/notebooks/astra_support.ipynb
```diff
@@ -19,18 +19,50 @@
   },
```
bjchambers, line 19: <pre class="ql-syntax" spellcheck="false">Line #29.        async for html in loader.alazy_load():
</pre><p>This is already simplified (the actual parsing of HTML links is extracted). I think a good next step (for future follow-up) would be a document transformer that does:</p><p> </p><ol><li>HTML to Markdown</li><li>With an optional content selector</li><li>With 0 or more "extractors to apply to beautiful soup".</li></ol><p> </p><p>This would let us replace this body with something like:</p><p> </p>

<pre class="ql-syntax" spellcheck="false">HtmlToMarkdownExtraction(
  selector = "...",
  soup_edge_extractors = [
    html_link_extractor,
  ]
)
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/datastax/ragstack-ai/pull/471/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

libs/knowledge-store/ragstack_knowledge_store/html_link_edge_extractor.py
```diff
@@ -0,0 +1,95 @@
...
+                 kind: str = "hyperlink",
+                 drop_fragments: bool = True):
+        """Extract hyperlinks from HTML content.
+
+        Expects the `page_content` to be HTML.
+
+        Args:
+            url_field: Name of the metadata field containing the URL
+                of the content. Defaults to "source".
+            kind: The kind of edge to extract. Defaults to "hyperink".
```
kerinin, line 61: `hyperlink`?

libs/knowledge-store/ragstack_knowledge_store/html_link_edge_extractor.py
```diff
@@ -0,0 +1,95 @@
...
+            input = BeautifulSoup(input, "html.parser")
+
+        url = document.metadata[self.url_field]
+        if self.drop_fragments:
+            url = urldefrag(url).url
+
+        hrefs = _parse_hrefs(input, url, self.drop_fragments)
+
+        link_tags = get_link_tags(document)
+        link_tags.add(IncomingLinkTag(kind=self._kind, tag=url))
```
kerinin, line 93: Maybe make this conditional on url not being `None`?

kerinin (approved):
LGTM

libs/knowledge-store/ragstack_knowledge_store/html_link_edge_extractor.py
bjchambers: I believe thiswould cause an error, unless they had done `doc.metadata["source"] = None`. Eg., the `document.metadata[self.url_field]` would raise `KeyError` if it isn't defined. And this is partly intentional -- without a URL, we can't correctly parse any relative links appearing in the content.

PR #480, opened 2024-06-11
[knowledge-store]: Use more generic types in API

cbornet:


libs/knowledge-store/ragstack_knowledge_store/base.py
```diff
@@ -113,13 +113,19 @@ async def aadd_nodes(
...
         """Add nodes to the knowledge store
 
         Args:
             nodes: the nodes to add.
         """
-        return await run_in_executor(None, self.add_nodes, nodes, **kwargs)
+        iterator = iter(await run_in_executor(None, self.add_nodes, nodes, **kwargs))
+        done = object()
+        while True:
+            doc = await run_in_executor(None, next, iterator, done)
```
bjchambers, line 125: Should this use `aiter` and `anext`?

libs/knowledge-store/ragstack_knowledge_store/cassandra.py
```diff
@@ -512,7 +512,7 @@ def similarity_search_by_vector(
     def _query_by_ids(
         self,
         ids: Iterable[str],
-    ) -> Iterable[Document]:
+    ) -> List[Document]:
```
bjchambers, line 515: Should these stick with `Iterable`? Wouldn't that let them change implementation more easily?

bjchambers (approved):

libs/knowledge-store/ragstack_knowledge_store/cassandra.py
cbornet: It's an internal method so the signature can be changed later if needed.

libs/knowledge-store/ragstack_knowledge_store/base.py
cbornet: `aiter` and `anext` work on `AsyncIterable`. Here `iterator` is an `Iterable`.

libs/knowledge-store/ragstack_knowledge_store/cassandra.py
cbornet: Yes, it 's better to keep `Iterable` to evolve it more easily.
Done.

PR #482, opened 2024-06-11
libs: add langflow package

nicoloboschi:


.github/workflows/ci-unit-tests.yml
```diff
@@ -138,9 +177,19 @@ jobs:
             rm -rf $dir/.tox
           }
 
-          run_itests libs/colbert
-          run_itests libs/langchain
-          run_itests libs/llamaindex
+          if [[ "true" == "${{ steps.filter.outputs.libs_langflow }}" ]]; then
```
cbornet, line 180: It doesn't seem to work: all integration tests were filtered. See https://github.com/datastax/ragstack-ai/actions/runs/9546095754/job/26308243791?pr=482#step:5:1

nicoloboschi: yep wrong syntax, thanks

PR #492, opened 2024-06-14
Rename KnowledgeStore to GraphStore

cbornet:


cbornet:
We should probably also rename the pypi package to `ragstack-ai-graph-store` and the root package to `ragstack_graph_store`.
WDYT @bjchambers @nicoloboschi ?

bjchambers:
> We should probably also rename the pypi package to `ragstack-ai-graph-store` and the root package to `ragstack_graph_store`. WDYT @bjchambers @nicoloboschi ?

Only reason to consider not is if we want to either (a) keep the core package separate in any way or (b) keep that name for other options. Maybe see where we land on naming? Not sure.

kerinin (approved):

PR #493, opened 2024-06-14
Update llama versions in langchain e2e tests

cbornet:


cbornet:
We also need a release of nemogardrails.

nicoloboschi (approved):

PR #497, opened 2024-06-17
deps: pin tenacity to != 8.4.0

nicoloboschi:


cbornet (approved):

cbornet (changes requested):
Also needed for e2e tests

PR #500, opened 2024-06-17
feat: Compute edges from tags rather than storing

bjchambers:
In dense graphs, there may be `O(node^2)` edges for tags with high connectivity. Storing the tags alone (`O(tags)`) allows for faster writes. We can further address dense graphs by using queries with similarity bounds and filters.

Other changes:

- refactor: Introduce `Link` as interface for `LinkTag` This allows non-tag based links to be added in the future.
- refactor: remove unused code
- fix notebook to use `ragstack_langchain` imports

Future improvements
- Further performance tuning of queries, including denormalization to allow for fewer queries during traversal.
- Consider / experiment with single key queries rather than `IN`.


libs/knowledge-store/ragstack_knowledge_store/concurrency.py
```diff
@@ -37,8 +37,9 @@ def _handle_result(
                 if self._pending == 0:
                     self._completion.notify()
 
-    def _handle_error(self, error):
+    def _handle_error(self, error, future: ResponseFuture):
         with self._completion:
+            print(f"Failed to execute {future.query}: {error}")
```
cbornet, line 42: Remove or use logger

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
```diff
@@ -118,7 +111,7 @@ def __init__(
         embedding: EmbeddingModel,
         *,
         node_table: str = "graph_nodes",
-        edge_table: str = "graph_edges",
+        targets_table: str = "graph_targets",
```
cbornet, line 114: This is a breaking change

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
```diff
@@ -215,47 +208,33 @@ def __init__(
...
-            SELECT content_id
+            SELECT link_to_tags
             FROM {keyspace}.{node_table}
-            WHERE link_to_tags CONTAINS ?
+            WHERE content_id IN ?
             """
         )
 
-        self._query_ids_and_embedding_by_link_to_tag = session.prepare(
+        # TODO: These queries require `ALLOW FILTERING` when run against Cassandra docker images.
```
cbornet, line 221: Using ALLOW FILTERING is extremely slow.
Can we do it with maybe partitionning ?

libs/langchain/ragstack_langchain/graph_store/cassandra.py
```diff
@@ -143,7 +127,7 @@ def similarity_search(
 
     def similarity_search_by_vector(
         self, embedding: List[float], k: int = 4, **kwargs: Any
-    ) -> List[Document]:
+    ) -> Iterable[Document]:
```
cbornet, line 130: I think we need to keep List as in the signature and change the code below.
I did it in the LangChain PR to be able to satisfy mypy.

libs/knowledge-store/ragstack_knowledge_store/links.py
```diff
@@ -35,20 +35,20 @@ def __init__(self, kind: str, tag: str) -> None:
         super().__init__(kind=kind, tag=tag, direction="bidir")
 
 
-LINK_TAGS = "link_tags"
+LINK_SET = "link_set"
 
 
-def get_link_tags(doc_or_md: Dict[str, Any]) -> Set[LinkTag]:
-    """Get the link-tag set from a document or metadata.
+def get_link_set(doc_or_md: Dict[str, Any]) -> Set[Link]:
```
cbornet, line 41: Maybe just `get_links` ?

libs/knowledge-store/ragstack_knowledge_store/links.py
```diff
@@ -35,20 +35,20 @@ def __init__(self, kind: str, tag: str) -> None:
         super().__init__(kind=kind, tag=tag, direction="bidir")
 
 
-LINK_TAGS = "link_tags"
+LINK_SET = "link_set"
```
cbornet, line 38: Maybe just `LINKS` ?

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
bjchambers: I think the ALLOW FILTERING isn't needed in Astra, but is in Cassandra. I can double check. I think the alternative is to blast out a bunch of separate queries rather than using the `IN`, which may actually be *faster*.

But, may be best to make that change as a follow-up so that we can get the bulk of the changes (including API changes) in. This would be a change under the hood, so to speak.

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
bjchambers: Yes.

libs/knowledge-store/ragstack_knowledge_store/links.py
bjchambers: Could be links. Wasn't sure if we wanted to try to keep it distinct from other things. I guess we could do something like `GRAPH_LINKS`... butI'd be ok with `LINKS`.

libs/langchain/ragstack_langchain/graph_store/base.py
```diff
@@ -460,9 +448,7 @@ async def _aget_relevant_documents(
         if self.search_type == "traversal":
             return [
                 doc
-                async for doc in self.vectorstore.atraversal_search(
-                    query, **self.search_kwargs
-                )
+                async for doc in self.vectorstore.atraversal_search(query, **self.search_kwargs)
```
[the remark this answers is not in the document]
bjchambers: `ruff`

libs/knowledge-store/ragstack_knowledge_store/links.py
kerinin: `get_links` seems more approachable - the fact that it's a set probably isn't that important.

libs/langchain/ragstack_langchain/graph_store/extractors/html_link_edge_extractor.py
```diff
@@ -99,7 +91,7 @@ def extract_one(
 
         hrefs = _parse_hrefs(input, url, self.drop_fragments)
 
-        link_tags = get_link_tags(document.metadata)
-        link_tags.add(IncomingLinkTag(kind=self._kind, tag=url))
+        link_set = get_link_set(document.metadata)
+        link_set.add(IncomingLinkTag(kind=self._kind, tag=url))
         for url in hrefs:
-            link_tags.add(OutgoingLinkTag(kind=self._kind, tag=url))
+            link_set.add(OutgoingLinkTag(kind=self._kind, tag=url))
```
kerinin, line 97: This is unrelated to this PR, but would it be possible to push this logic into the Document? From a usability perspective I feel like the best  API would be something like

```python
document.link_to(kind=..., tag=...)
document.link_from(...)
document.link_bidir(...)
```

I suspect that would require workarounds we don't like (subclassing `Document` or monkey-patching or similar) but maybe it's possible?

Alternatively, I would probably prefer mutating methods, like

```python
from ragstack_knowledge_store.links import link_to, link_from, link_bidir

link_to(document, kind=..., tag=...)
# etc
```

The thing I dislike about the example above is that it doesn't look like the document is actually being modified - `get_link_set` returns a value which is implicitly shared with the document, and that just feels wrong to me. It also introduces the concept of a "link set", which makes it harder to explain what's happening (I say this from experience trying to explain code samples to several people). A casual user shouldn't need to understand how the graph metadata is stored - they just want to create edges. A mutating function is more clear in that it's clear what you're doing (linking things together), what you're operating on (the document) and how you're modifying it (kind, tags, etc).

libs/langchain/ragstack_langchain/graph_store/base.py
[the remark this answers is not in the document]
bjchambers: Ruff formatting is compatible with Black, IIUC: 

> ⚖️ Drop-in parity with Flake8, isort, and Black

libs/langchain/ragstack_langchain/graph_store/cassandra.py
bjchambers: I don't think this is correct then -- you can't return a `List[Document]` and use `yield` inside the method. `yield` makes it a generator, not a `List`. Specifically, the type annotation here is wrong:

https://github.com/datastax/ragstack-ai/blob/main/libs/langchain/ragstack_langchain/graph_store/cassandra.py#L148

bjchambers: Done, but changed to actually return a list.

libs/langchain/ragstack_langchain/graph_store/extractors/html_link_edge_extractor.py
bjchambers: That was introduced in a previous PR, so definitely unrelated to this. I don't think that formulation is ideal since it requires more access via the metadata.

Unfortunately, adding this for document is a bit tricky -- methods on Document would be ugly to add.

I think the best option is:

```
def add_links(doc_or_md: Union["langchain_core.documents.Document", Dict[str, Any]], *links: Link) -> None:
    ...
```

Note it has to do some weird typing, since the `Document` import may not be available -- this may clear up when @cbornet moves this around as part of moving to langchain.

kerinin: 👍 `add_links` seems like an improvement.

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
bjchambers: Yeah. Different schema too.

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
bjchambers: Done.

libs/langchain/ragstack_langchain/graph_store/base.py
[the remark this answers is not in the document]
cbornet: Then it's surprising that the formatting was changed here...

libs/knowledge-store/ragstack_knowledge_store/links.py
```diff
@@ -0,0 +1,78 @@
...
+    """Get the links from a document or metadata.
+
+    Args:
+        doc_or_md: The metadata to get the link tags from.
+
+    Returns:
+        The set of link tags from the document or metadata.
+    """
+    try:
+        from langchain_core.documents import Document
```
cbornet, line 55: I don't see the advantage of supporting Document in get_links.
It's easy to do `get_links(doc.metadata)`

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
cbornet: Then we need to cut the release with this change before the LangChain PR is merged.

libs/knowledge-store/ragstack_knowledge_store/links.py
bjchambers: It's not a dependency -- it's optional. If used from a project with Langchain it lets the user write `add_links(doc, x, y, z)` and more clearly conveys "add these links to the document" rather than `add_links(doc.metadata, x, y, z)`. I'm fine dropping this, but it seemed beneficial to let the user think about this as "adding links to the document".

kerinin: I feel like accepting documents here helps  communicate the API and makes for cleaner examples. Loose typing is pretty standard  in Python, for better or worse

libs/knowledge-store/ragstack_knowledge_store/links.py
```diff
@@ -0,0 +1,78 @@
...
+        super().__init__(kind=kind, tag=tag, direction="incoming")
+
+
+@dataclass(frozen=True)
+class BidirLinkTag(LinkTag):
+    def __init__(self, kind: str, tag: str) -> None:
+        super().__init__(kind=kind, tag=tag, direction="bidir")
+
+
+LINKS = "links"
```
cbornet, line 42: I called it METADATA_LINKS_KEY in the other PR.

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
bjchambers: Generally agree we should get this in (as well as the one I have started on after this to actually serialize metadata) since they are schema changes. That said, I don't know that we won't be able to make these changes after the LangChain PR is merged -- changes will be necessary, and we can't treat that as a point after which we can never make changes.

libs/knowledge-store/ragstack_knowledge_store/links.py
cbornet: My feeling is that accepting 2 different types of input doesn’t make it clearer. You have to explain in comments that it’s « doc or metadata » and generally I would advocate for having 2 methods.
Anyway, not a hill I would die on. So it can be left that way.

bjchambers: I'm going with this for the moment. I *think* I could be convinced the method should just move into `langchain` or the `langchain` package, at which point, the dependency on `Document` makes sense -- and maybe we drop support for metadata.

cbornet: Yes it should at least be moved to the langchain package.

PR #503, opened 2024-06-18
adding MimeNode type

mieslep:


libs/knowledge-store/ragstack_knowledge_store/embedding_model.py
```diff
@@ -1,22 +1,67 @@
...
     """Embedding model."""
 
-    @abstractmethod
+    def __init__(self, embeddings: Any, method_map: Optional[dict] = None, other_methods: Optional[List[str]] = None):
+        self.embeddings = embeddings
+        self.method_name = {}
+        method_map = method_map if method_map else {}
+        other_methods = other_methods if other_methods else []
+
+        base_methods = ['embed_texts', 'aembed_texts', 'embed_query', 'aembed_query']
```
bjchambers, line 13: I don't think we should try to add all of these as methods, it's definitely pretty messy.

I think we should just have `embed_mime(self, mime_type: str, content: Union[str, Bytes])` or something like that. Then there is only a single abstract method to use for any mime type and the names can be different, etc.

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
```diff
@@ -326,52 +335,71 @@ def add_nodes(
...
+                    for node in nodes_list:
+                        embedding = self._embedding.invoke(singular_method_name, node.text if isinstance(node, TextNode) else node.content)
+                        embeddings.append(embedding)
+                    embeddings_dict[mime_type] = embeddings
+                else:
+                    raise NotImplementedError(f"No embedding method available for MIME type: {mime_type}, implemented methods: {self._embedding.implements()}.")
+
 
         # Step 1: Add the nodes, collecting the tags and new sources / targets.
+        tag_to_new_sources = {}
```
bjchambers, line 377: This has changed significantly from the previous implementation. I think it will need to be reworked if we work on adding this right now.

libs/knowledge-store/ragstack_knowledge_store/embedding_model.py
mieslep: 100% but right now, LangChain doesn't have "embed_mime" :)

cbornet:
Shall we close this one ?

PR #507, opened 2024-06-19
Add Links to graph store Node

cbornet:


libs/knowledge-store/ragstack_knowledge_store/graph_store.py
```diff
@@ -18,7 +19,7 @@
 from .concurrency import ConcurrentQueries
 from .content import Kind
 from .embedding_model import EmbeddingModel
-from .link_tag import get_link_tags
+from .link_tag import LinkTag
```
bjchambers, line 22: Suggest changing this to just `links` and `Link` as well. The original name (`LinkTag`) only supported using tags. But, I anticpate other kinds of links (for instance, we could use a special embedding model, and have a link if the source embeds near the target). So, that would be a `OutgoingEmbeddingLink` or something.

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
```diff
@@ -31,8 +32,9 @@ class Node:
     id: Optional[str] = None
     """Unique ID for the node. Will be generated by the GraphStore if not set."""
     metadata: dict = field(default_factory=dict)
-    """Metadata for the node. May contain information used to link this node 
-    with other nodes."""
+    """Metadata for the node."""
+    links: Set[LinkTag] = field(default_factory=set)
```
bjchambers, line 36: This should be `Links` so that it can be tags and other kinds of links.

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
```diff
@@ -328,11 +330,13 @@ def add_nodes(
     ) -> Iterable[str]:
         texts = []
         metadatas = []
+        links: List[Set[LinkTag]] = []
```
bjchambers, line 333: `Link`

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
```diff
@@ -353,7 +357,7 @@ def add_nodes(
                 link_to_tags = set()  # link to these tags
                 link_from_tags = set()  # link from these tags
 
-                for tag in get_link_tags(metadata):
+                for tag in _links:
```
bjchambers, line 360: Why underscore? I usually read that as unused? If this is to avoid shadowing, I think we could call things `link_sets` and `link_set` in this method.

libs/langchain/ragstack_langchain/graph_store/base.py
```diff
@@ -36,8 +39,9 @@ class Node(Serializable):
     id: Optional[str]
     """Unique ID for the node. Will be generated by the GraphStore if not set."""
     metadata: dict = Field(default_factory=dict)
-    """Metadata for the node. May contain information used to link this node
-    with other nodes."""
+    """Metadata for the node."""
+    links: Set[LinkTag] = Field(default_factory=set)
```
bjchambers, line 43: `Link` not `LinkTag`?

libs/langchain/ragstack_langchain/graph_store/extractors/edge_extractor.py
```diff
@@ -1,28 +1,25 @@
...
 from ragstack_knowledge_store._utils import strict_zip
-from ragstack_knowledge_store.link_tag import LinkTag
 
 InputT = TypeVar("InputT")
 
 
 class EdgeExtractor(ABC, Generic[InputT]):
     @abstractmethod
-    def extract_one(self, document: Document, input: InputT):
+    def extract_one(self, document: Document, input: InputT) -> None:
```
bjchambers, line 14: I wonder if it would be cleaner for this to not take `Document` and return the `Set[Link]` instead. Then we could allow: `add_links(document, link1, link2, link3)` or `add_links(document, *edge_extractor(input))`, etc.

libs/langchain/ragstack_langchain/graph_store/extractors/html_link_edge_extractor.py
```diff
@@ -1,6 +1,6 @@
 from langchain_core.documents import Document
-from ragstack_knowledge_store.link_tag import (
-    get_link_tags,
+from ragstack_langchain.graph_store.links import (
```
bjchambers, line 2: I think it may be worth putting the edge extractor interface in the common code so we could share it with LlamaIndex, etc. That may be enabled by the EdgeExtractor change.

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
cbornet: OK, I'll do the same as in your PR. Whichever one we merge first.

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
cbornet: Underscore-something is not the convention for unused, rather for private args.
But I agree we can rename better here.

libs/langchain/ragstack_langchain/graph_store/base.py
cbornet: Good catch

libs/langchain/ragstack_langchain/graph_store/extractors/edge_extractor.py
cbornet: I also prefer not mutating args when the function name is not clearly explicit about it.

libs/langchain/ragstack_langchain/graph_store/extractors/html_link_edge_extractor.py
cbornet: It's always better to make it more generic.
Howver if we want to upstream extractors, I think we shouldn't have a dependency on  ragstack for them.
So LangChain and Llama-index would have their own copy/implementation of them.
WDYT ?

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
cbornet: done

libs/langchain/ragstack_langchain/graph_store/extractors/edge_extractor.py
bjchambers: Yeah. I think I like that. We could change

```
def add_links(doc_or_md, *links: Link) -> None:
  ...
```

to
```
def add_links(doc_or_md, *links: Union[Link, Iterable[Link]]) -> None:
  ...
```

To make it even easier. Then we can write `add_links(document, edge_extractor(input), edge_extractor2(input))` and it will automatically add all the edges to the document or metadata. And the `add_links` becomes the only thing that mutates the document.

cbornet: Let's o another PR for that.

libs/langchain/ragstack_langchain/graph_store/extractors/html_link_edge_extractor.py
cbornet: Let's see that as a follow-up

PR #514, opened 2024-06-24
feat: Limit items retrieved during MMR traversal

bjchambers:


libs/knowledge-store/ragstack_knowledge_store/graph_store.py
```diff
@@ -559,6 +582,8 @@ def similarity_search(
     def _get_adjacent(
```
bjchambers, line 582: Annoyingly, `_get_adjacent` is also used in the testing method `_linked_ids`. I propose we get rid of it. Then I wouldn't need the version without embedings, since it is only used by MMR.

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
```diff
@@ -590,6 +623,7 @@ def add_targets(rows):
             for source_id in source_ids:
                 cq.execute(self._query_source_tags_by_id, (source_id,), callback=add_sources)
 
+        # TODO: Consider a combined limit based on the similarity and/or predicated MMR score?
```
bjchambers, line 626: Right now, each tag can contribute up to `k_per_tag` items. Conceivably, we could try to bound this to an *actual* limit. Note that retrieving more here won't lead to picking more actual items, since the MMR search will still be limited. So, at most this would let us limit the growth of the tags being considered. This may be worth considering.

It gets a bit harder / weirder when we talk about other kinds of edges. Specifically, if we used a source-to-target embedding, then we'd be relying on the ANN ordering to choose the "N nearest targets" to each source tag. At that point, we wouldn't be filtering to the "N nearest based on the source similarity". So, we my thinking is maybe:

1. Call this `k_per_tag` -- it is a limit of how many items we retrieve per tag.
2. For equality tags, we retrieve `k_per_tag` ordered by target similarity to the question.
3. For embedding tags, we retrieve `k_per_tag` ordered by the target embedding (near to the source).
4. For everything we retrieve, we populate a list and scale it based on candidates.
5. We only return the top `k_traverse` (based on the MMR score computed from similarity and diversity).

This let's us say `k_per_tag` is a limit on the number of rows fetched *for each tag* and `k_traverse` is a limit on how many items each selected node can add to the set being considered by MMR.

bjchambers: Thinking about this some more I think the best options are:

1. For both equality and embedding links to query for the top `fetch_k_per_tag` and compute their scores.
2. Add a bound on the size of the set being considered, and drop the lowest scored items if it exceeds that.

This will provide the most consistent behavior between tags, and the most understandable limit in terms of size of the queue. This can be added in a future change -- the current change should be good with the existing `k_per_tag` behavior.

cbornet (approved):

PR #516, opened 2024-06-25
ref: Rename EdgeExtractor to LinkExtractor

bjchambers:
This more closely matches how these operate. For each node we extract "links" (roughly, ports) indicating incoming/outgoing connections they should have.

Edges are created while traversing based on the links. A node with an outgoing link has edges to nodes with matching incoming links.

Also adds tests for the HTML link extractor.


libs/langchain/ragstack_langchain/graph_store/extractors/link_extractor.py
```diff
@@ -0,0 +1,48 @@
...
+            input: The input content to extract edges from.
+            **kwargs: Additional keyword arguments for the extractor.
+
+        Returns:
+            Set of links extracted from the input.
+        """
+
+    def extract_batch(self,
+                      inputs: Iterable[InputT],
+                      batch_kwargs: Optional[Iterable[Dict[str, Any]]] = None,
```
bjchambers, line 34: This was a bit ugly. The problem is that extractors may need additional information about the input -- for instance, HTML Link Extractor needs to know the Base URL. The thinking here is that it can be passed in via extra arguments.

An alternative, would be to have it set `InputT = Tuple[str, str]` (basically, take a tuple of the URL and the input). That may be cleaner (only one iterable). Thoughts?

libs/knowledge-store/notebooks/astra_support.ipynb
```diff
@@ -125,10 +105,11 @@
     "from typing import AsyncIterator, Iterable\n",
     "from ragstack_knowledge_store.graph_store import CONTENT_ID\n",
     "from markdownify import MarkdownConverter\n",
-    "from ragstack_langchain.graph_store.extractors import HtmlLinkEdgeExtractor\n",
+    "from ragstack_langchain.graph_store.links import add_links\n",
```
bjchambers, line 108: I wonder if we should tinker with the `__init__.py` so this can be `from ragstack_langchain.graph_store import add_links` or even make it part of `extractors`?

libs/langchain/ragstack_langchain/graph_store/extractors/link_extractor.py
```diff
@@ -0,0 +1,48 @@
...
+
+        Args:
+            input: The input content to extract edges from.
+            **kwargs: Additional keyword arguments for the extractor.
+
+        Returns:
+            Set of links extracted from the input.
+        """
+
+    def extract_batch(self,
```
cbornet, line 32: If we have `extract_one`, then maybe this should be `extact_many` ?

libs/langchain/ragstack_langchain/graph_store/extractors/link_extractor.py
cbornet: > An alternative, would be to have it set InputT = Tuple[str, str] 

I prefer this solution. And maybe InputT could be typed stronger: InpuT = HtmlLinkInput

libs/knowledge-store/notebooks/astra_support.ipynb
bjchambers: @cbornet thoughts on naming or just moving `add_links` into extractors?

cbornet: Yes, I think we can put it in `__init__`.

libs/langchain/ragstack_langchain/graph_store/extractors/html_link_extractor.py
```diff
@@ -71,32 +74,29 @@ def __init__(
...
+        hrefs = _parse_hrefs(content, base_url, self.drop_fragments)
 
-        add_links(
-            document,
-            Link.incoming(kind=self._kind, tag=url),
-            *[Link.outgoing(kind=self._kind, tag=url) for url in hrefs]
-        )
+        links = { Link.outgoing(kind=self._kind, tag=url) for url in hrefs }
+        links.add(Link.incoming(kind=self._kind, tag=base_url))
+        return links
```
cbornet, line 102: Can you run black on this code ?
We should have an automatic check for it.

cbornet (approved):

PR #518, opened 2024-06-26
Extract metadata's content_id to node id

cbornet:


libs/knowledge-store/ragstack_knowledge_store/graph_store.py
```diff
@@ -24,41 +24,38 @@
...
-CONTENT_ID = "content_id"
-
 
 @dataclass
 class Node:
     """Node in the GraphStore"""
 
     id: Optional[str] = None
     """Unique ID for the node. Will be generated by the GraphStore if not set."""
+    text: str = None
```
bjchambers, line 34: I wonder if we should call this `content`. In the future, we could change it to `Union[str, bytes]` to support different mime types with inline images or not, etc. That said, I think that can be evolved later if needed.

libs/langchain/ragstack_langchain/graph_store/base.py
```diff
@@ -245,14 +248,16 @@ async def atraversal_search(
...
+                None, self.traversal_search, query, k=k, depth=depth, **kwargs
+            )
         )
         done = object()
         while True:
             doc = await run_in_executor(None, next, iterator, done)
             if doc is done:
                 break
-            yield doc
+            yield doc  # type: ignore[misc]
```
bjchambers, line 260: Curious what the lint is complaining about? What are we suppressing here?

bjchambers (approved):

libs/langchain/ragstack_langchain/graph_store/base.py
cbornet: I'm aligning with the langchain code so it's easier to port changes.
This is to appease mypy.
Basically because run_in_executor messes the types.

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
cbornet: Yes we can see that later.

PR #527, opened 2024-06-26
feat: Add extractors for keywords, ner & hierarchy

bjchambers:
This simplifies the use of Keybert for keyword based links and GLiNER for named-entity based links.

This also makes it easier to create links representing a document and/or page hierarchy.


cbornet:
That's a good question between: do we have a ragstack lib that people can use as an optional dependency in LangChain or Llama-index ? Then in LangChain we need to create components to use it.
Or do we put this duplicate this code in LangChain and Llama-Index and make it compatible with their respective abstractions ?

kerinin:
That makes sense. 

FWIW, I think the extractors assign a `source` metadata field which is generally the URL path for remote pages. Capturing the HREF's would probably depend on how the HTML is parsed though.

bjchambers:
@kerinin Take a look at the new tests in the latest commit. I think this strikes a good balance -- if you care about how many times the HTML is parsed, etc. you can use the link extractors directly. If you want to use document transformers, you can do that. In theory, we could even create other document transformers that do things like "parse HTML, then select the sub-parts, then apply extracts to the beautifulsoup, then convert to markdown", but don't want to let the cases explode. This demonstrates an API that supports documents, and can be extended.

kerinin (commented):
Looks great!

PR #528, opened 2024-06-27
initial add of ragulate to ragstack

epinzur:
code taken from https://github.com/epinzur/ragulate


epinzur:
@nicoloboschi maybe we merge the code first and then work on CI & releasing separately?

nicoloboschi (approved):

PR #529, opened 2024-06-27
test: Restore knowldege-store integration tests

bjchambers:
These currently just demonstrate that the knowledge store can be created. That is enough to verify the schema and queries are correct. I'll add more tests over time.

These are smaller than the full langchain tests to run, so are easier to use for iterating on the details.


libs/knowledge-store/tests/integration_tests/test_graph_store.py
```diff
@@ -0,0 +1,79 @@
...
+            session = session,
+            keyspace = KEYSPACE,
+            node_table = node_table,
+            targets_table = targets_table,
+        )
+
+    yield _make_graph_store
+
+    for table in tables:
+        session.execute(f"DROP TABLE IF EXISTS {table}")
```
bjchambers, line 70: Arguably, since these only run against a cassandra docker image, I could skip dropping things for even faster tests. Thoughts?

nicoloboschi: yes you could indeed. Would be great to refactor this to depends on ragstack-test-utils which already have the container code

bjchambers: I would, but I'm worried that will lead to the same problems I have with `libs/langchain/tests/integration_tests`. Specifically, I can't run those locally because something just hangs forever about starting up the container. This works quickly and is easy to interact with since it's all in one file. I generally prefer tests being straightforward, and this seems to simplify things (keeping it in one place) rather than entangling with other utilities...

libs/tests-utils/ragstack_tests_utils/cassandra_container.py
```diff
@@ -12,6 +12,14 @@ def __init__(
         super(CassandraContainer, self).__init__(image=image, **kwargs)
         self.port = port
 
+        self.with_env(
```
bjchambers, line 15: @nicoloboschi 

At least the `HEAP_NEWSIZE` and `MAX_HEAP_SIZE` were why this container wasn't starting for me locally. With these changes things work -- both `knowledge-store` and `langchain`.

Note for Langchain, we can run `poetry run pytest tests/integration_tests -k "[cassandra]` to run only those tests that use the local container (not sure if there is a way to make this a default, so it's easier to run without the astra environment setup).

PR #531, opened 2024-06-28
ragulate: ci and release process

nicoloboschi:


.gitignore
```diff
@@ -100,6 +100,7 @@ ipython_config.py
 #   commonly ignored for libraries.
 #   https://python-poetry.org/docs/basic-usage/#commit-your-poetrylock-file-to-version-control
 poetry.lock
+!libs/ragulate/poetry.lock
```
epinzur, line 103: why ignore this? I've specifically added it to the repo to control versions of things.

epinzur: eh, this is to not ignore... i get it.

epinzur (approved):

libs/ragulate/ragstack_ragulate/datasets/utils.py
```diff
@@ -1,17 +1,19 @@
...
+DATASET_ROOT_PATH = dirname(abspath(__file__))
+
 
 def find_dataset(name: str) -> BaseDataset:
-    root_path = "datasets"
     name = inflection.underscore(name)
-    for kind in os.listdir(root_path):
-        kind_path = os.path.join(root_path, kind)
+    for kind in os.listdir(DATASET_ROOT_PATH):
+        kind_path = os.path.join(DATASET_ROOT_PATH, kind)
```
epinzur, line 16: actually it would be good if we can keep all the datasets under some sub-path.

libs/ragulate/ragstack_ragulate/datasets/utils.py
```diff
@@ -1,17 +1,19 @@
 import os
+from os.path import dirname, abspath
 
 import inflection
 
 from .base_dataset import BaseDataset
 from .crag_dataset import CragDataset
 from .llama_dataset import LlamaDataset
 
+DATASET_ROOT_PATH = dirname(abspath(__file__))
```
epinzur, line 10: this is incorrect. it is not the path of this code folder.  it is a folder in the user's directory to store datasets.

I'll revert this change and make a fix for when it doesn't exist.

PR #536, opened 2024-06-28
ci: run ruff check

nicoloboschi:


libs/ragulate/ragstack_ragulate/cli.py
```diff
@@ -1,5 +1,4 @@
 # this must be imported first for dynamic module loading to function properly
-import setuptools  # isort: skip
```
epinzur, line 2: we need to keep this one

epinzur: @nicoloboschi there is prob a way for ruff to ignore lines

nicoloboschi: added back

cbornet (approved):

tox.ini
```diff
@@ -47,18 +47,19 @@ commands =
 description = lint
 skip_install = true
 commands =
-    poetry install --no-root --only lint
-    poetry run ruff format --diff ragstack tests
-    poetry run ruff check --select I ragstack tests
-    poetry run ruff check ragstack tests
-    poetry run yamllint -s -c .github/.yamllint .github/
+    ruff format --diff .
```
cbornet, line 50: Please keep the changes from main

cbornet (changes requested):

cbornet (approved):

PR #552, opened 2024-06-30
Fix Resume RAGulate Queries

saikolasani:
preset the query count (_finished_queries) to the count of completed queries and update tqdm


epinzur (approved):

saikolasani:
@jordanrfrazier Tagged for visibility

PR #556, opened 2024-07-02
Add ruff rules for pylint

cbornet:


libs/colbert/tests/integration_tests/test_database.py
```diff
@@ -51,7 +51,7 @@ def test_database_sync(request, vector_store: str):
 
     results = database.add_chunks(chunks=[chunk_0, chunk_1])
 
-    assert len(results) == 2
+    assert len(results) == 2  # noqa: PLR2004
```
cbornet, line 54: PLR2004 is "no magic numbers rule". Not always relevant.

bjchambers: It may be worth having a separate config for the test directories? Then we could omit some of the lints that are less relevant to tests?

https://github.com/astral-sh/ruff/discussions/4119

cbornet: For the moment I'm using pre-file-ignores to ignore some rules for tests. But yes, it could be a separate config.

PR #558, opened 2024-07-02
Add ruff-specific rules (RUF)

cbornet:


libs/ragulate/ragstack_ragulate/pipelines/query_pipeline.py
```diff
@@ -30,8 +30,8 @@ class QueryPipeline(BasePipeline):
     _tru: Tru
     _name: str
     _progress: tqdm
-    _queries: Dict[str, List[str]] = {}
-    _golden_sets: Dict[str, List[Dict[str, str]]] = {}
+    _queries: Dict[str, List[str]]
```
cbornet, line 33: This was an actual bug !

libs/knowledge-graph/ragstack_knowledge_graph/traverse.py
```diff
@@ -6,10 +6,19 @@
 from cassio.config import check_resolve_keyspace, check_resolve_session
 
 
-class Node(NamedTuple):
+class _Node(NamedTuple):
     name: str
     type: str
-    properties: Dict[str, Any] = {}
+    properties: Dict[str, Any]
```
cbornet, line 12: This was an actual bug !

libs/ragulate/ragstack_ragulate/pipelines/query_pipeline.py
epinzur: can you explain how this is a bug?

PR #562, opened 2024-07-03
Add various ruff rules

cbornet:


libs/e2e-tests/e2e_tests/langchain/test_unstructured.py
```diff
@@ -15,9 +17,9 @@
 )
 from e2e_tests.langchain.rag_application import BASIC_QA_PROMPT
 from e2e_tests.test_utils import get_local_resource_path
-from e2e_tests.test_utils.vector_store_handler import (
-    VectorStoreTestContext,
-)
+
```
cbornet, line 20: Fixes TCH001: https://docs.astral.sh/ruff/rules/typing-only-first-party-import/

PR #565, opened 2024-07-03
Add ruff rules for builtins

cbornet:


libs/langchain/ragstack_langchain/graph_store/extractors/gliner_link_extractor.py
```diff
@@ -39,7 +39,7 @@ def __init__(
         self._kind = kind
         self._extract_kwargs = extract_kwargs or {}
 
-    def extract_one(self, input: GLiNERInput) -> Set[Link]:
+    def extract_one(self, input: GLiNERInput) -> Set[Link]:  # noqa: A002
```
cbornet, line 42: @bjchambers should we use another param name than `input` to prevent shadowing ?

bjchambers: I'm somewhat OK shadowing `input`. Is there some background/context for why we want to add this rule?

cbornet: That’s a common rule in Python .
Pycharm even warns for it.
It can be bothersome when you can’t use a builtin because it has been shadowed.

PR #567, opened 2024-07-03
[Docs] - Update Colbert doc to reflect notebook updates

mendonk:
* Update retriever method
* Offer install as extra as an option
* Organize imports and steps a little more intuitively
* Code tested and works


docs/modules/examples/pages/colbert.adoc
```diff
@@ -15,48 +15,55 @@ Use ColBERT, Astra DB, and RAGStack to:
...
-pip install ragstack-ai-colbert
+pip install ragstack-ai-colbert python-dotenv
+----
++
+. Create a `.env` file in your application directory with the following environment variables:
++
+[source,bash]
+----
+ASTRA_DB_APPLICATION_TOKEN=AstraCS: ...
+ASTRA_DB_ID=2eab82dc-9032-45ba-aeb0-a481b6f9458d
```
epinzur, line 37: maybe don't use an actual DB ID but show an example one? your-db-id-here

epinzur (approved):

docs/modules/examples/pages/colbert.adoc
mendonk: It's a fake string

PR #577, opened 2024-07-06
Add type checking of test-utils and knowledge-store

cbornet:


libs/knowledge-store/ragstack_knowledge_store/graph_store.py
```diff
@@ -351,16 +350,15 @@ def _nodes_with_ids(
...
         with self._concurrent_queries() as cq:
 
-            def add_nodes(rows):
+            def add_nodes(rows: Iterable[Any]) -> None:
                 for row in rows:
                     results[row.content_id] = _row_to_node(row)
 
             for node_id in ids:
                 if node_id not in results:
-                    results[node_id] = None
```
cbornet, line 363: @bjchambers this complicates the typing and doesn't seem needed. So I removed it. OK for this ?

bjchambers: We may want this, since it changes the behavior of the function on duplicate IDs. We need to either verify we won't call it with duplicates, or document that it will issue multiple requests in that case. Or, we should leave as is. I don't think it should complicate the typing excessively. I would suggest:

1. `results: Dict[str, Optional[Node]]`
2. `return [results[node_id] for node_id in ids if results[node_id] is not None]`. This could also be an assertion, since it being `None` means we requested an invalid ID. Alternatively, we may want to handle invalid ID in the `add_nodes` case (an empty `rows` indicates a problem).

The result type should still be `Node`.

libs/knowledge-store/tests/integration_tests/test_graph_store.py
```diff
@@ -21,14 +21,30 @@ def cassandra() -> Iterator[LocalCassandraTestStore]:
         store.docker_container.stop()
 
 
+class DummyEmbeddingModel(EmbeddingModel):
+    def embed_texts(self, _: List[str]) -> List[List[float]]:
+        return []
```
bjchambers, line 26: This won't work -- returning empty embeddings is invalid. In a PR I'm working on I added a dependency on `ragstack-langchain` so this could use the adapter there.

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
cbornet: I've put it back. The handling of None values (or empty resultset in the callback) can be done in a subsequent PR.

libs/knowledge-store/tests/integration_tests/test_graph_store.py
cbornet: I don't think it's a good thing to have a dependency on ragstack-langchain. That's a circular dependency...
For the test, we only need embed_query to return a non-zero length vector so I return a fixed one.

cbornet: OK. We can use OpenAI or HuggingFace later when we need it. Keeping the test dependency on ragstack-ai-langchain for now.

cbornet: We'll probably remove the graph store from ragstack-ai-langchain at some point.
So for tests I think it would probably be better to use langchain directly.

PR #578, opened 2024-07-08
suggestion

bjchambers:


libs/knowledge-store/ragstack_knowledge_store/graph_store.py
```diff
@@ -350,20 +350,32 @@ def _nodes_with_ids(
...
     ) -> List[Node]:
-        results: Dict[str, Node] = {}
+        results: Dict[str, Optional[Node]] = {}
         with self._concurrent_queries() as cq:
 
             def add_nodes(rows: Iterable[Any]) -> None:
+                # Should always be exactly one row here. We don't need to check
+                #   1. The query is for a `ID == ?` query on the primary key.
+                #   2. If it doesn't exist, the `get_result` method below will
+                #      raise an exception indicating the ID doesn't exist.
```
cbornet, line 360: Can't we raise the exception here if len(rows) != 1 ?

cbornet:
Merged

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
bjchambers: We could. That raised questions: if we get `len(rows)` will we be able to iterate it again? if we wish to report an error, we also need to pass the `id` in, etc. Figured this was enough for now.

PR #579, opened 2024-07-09
[Ragulate] Save Results + Update Model Providers

saikolasani:
When query finishes, output the raw data in json format. Also removes bedrock, langchain, and litellm as providers. They require further work to be done later.


saikolasani:
Tagging @jordanrfrazier for visibility

libs/ragulate/ragstack_ragulate/pipelines/query_pipeline.py
```diff
@@ -142,6 +144,7 @@ def stop_evaluation(self, loc: str):
                 logger.exception("issue stopping evaluator")
             finally:
                 self._progress.close()
+                self.export_results()
```
epinzur, line 147: I think this will output both when the processing finishes when the run is complete, and when someone presses Ctrl-C to cancel in the middle.

Can we only output when the processing is actually finished?

jordanrfrazier: Would it be possible to output the results at the stopped time (would you want to see intermediate results just to check?) and then overwrite or append-only when processing is finished?

epinzur: I think overwrite would be fine after finsihed.

I'm actually thinking that this might be better as a separate command though.  Like maybe `export`? 

Because for many cases, people probably won't be using these data files... just the charts.

epinzur: we can just merge this for now and see how we like the behavior or if we want to move it to a separate command.

epinzur (approved):

PR #580, opened 2024-07-09
enabled metadata support in ragulate

epinzur:


libs/ragulate/ragstack_ragulate/datasets/base_dataset.py
```diff
@@ -3,24 +3,37 @@
...
+from typing import Any, Dict, List, Optional
 
 import aiofiles
 import aiohttp
 from tqdm.asyncio import tqdm
 
 
+class QueryItem():
+    query: str
+    metadata: Dict[str, Any]
```
cbornet, line 15: Need to rebase with new ruff rules

cbornet (changes requested):

PR #584, opened 2024-07-10
Bump langchain versions

cbornet:


libs/langchain/ragstack_langchain/graph_store/base.py
```diff
@@ -1,531 +1,12 @@
...
-            ]
-        if self.search_type == "mmr_traversal":
-            return [
-                doc
-                async for doc in self.vectorstore.ammr_traversal_search(
-                    query, **self.search_kwargs
-                )
-            ]
-        return await super()._aget_relevant_documents(query, run_manager=run_manager)
+__all__ = ["GraphStore", "GraphStoreRetriever", "Node", "nodes_to_documents"]
```
bjchambers, line 12: I don't know that it is necessary to preserve backwards compatibility. I think it would be better to get to the new names sooner. To that end, we may want to change the names to the `GraphVector*` sooner rather than later. But, that can happen in a follow-up PR since it will likely need to change notebooks, etc.

libs/langchain/tests/integration_tests/test_graph_store.py
```diff
@@ -1,283 +0,0 @@
-import math
```
bjchambers, line 1: I'm a bit worried that we're deleting these tests. Are we sure they're running in LangChain? There is some benefit to keeping them here (I think) -- ideally, we'd have a way of running tests on the VectorGraphStore when we change `knowledge-store` that uses that rather than the released version...

libs/langchain/tests/unit_tests/test_graph_store.py
```diff
@@ -1,65 +0,0 @@
...
-from langchain_core.documents import Document
-from ragstack_langchain.graph_store.base import (
-    Node,
-    _documents_to_nodes,
-    _texts_to_nodes,
-)
-from ragstack_langchain.graph_store.links import Link
-
-
-def test_texts_to_nodes() -> None:
```
bjchambers, line 11: Same (although I'm more convinced these running in LangChain is fine, since most of this logic is in LangChain).

libs/langchain/tests/integration_tests/test_graph_store.py
cbornet: I think we should have other tests. Maybe in e2e-tests project.

libs/langchain/tests/unit_tests/test_graph_store.py
cbornet: These are clearly for LangChain only

libs/langchain/ragstack_langchain/graph_store/base.py
cbornet: I think people who want to upgrade should not just change GraphStore to GraphVectorStore but use langchain instead of ragstack.
I'll update the notebooks once we have a release (since we are still usingthe extractors from ragstack-langchain and we need a compatible langchain version)

cbornet: I don't know if we need a deprecation warning system.
Probably not a lot of people are using this atm and this is not really needed. Also it's easy to upgrade.

libs/langchain/tests/integration_tests/test_graph_store.py
cbornet: > Are we sure they're running in LangChain?

Indeed they're not. Maybe we'll keep them here for the moment. And move them to e2e-tests later.

cbornet: restored.

libs/langchain/ragstack_langchain/graph_store/base.py
bjchambers: That makes sense. In the long term, do we see `ragstack_langchain/graph_store` going away, or do we think there will be utility in us providing extensions (eg., additional extractors, etc.) that are part of our package? If the latter, do we think it's worth always re-exporting things so the extensions can be used in one place? Or only exporting the extensions to emphasize that everyhting else comes from LangChain?

bjchambers (approved):

libs/langchain/ragstack_langchain/graph_store/base.py
cbornet: Yes, in the long term, we can have components that stay into RAGStack.
But for the components that are in LangChain, I think we will eventually remove the reexports.

cbornet:
There is a breaking change due to the removal of target_tables.
We need to adapt LangChain first.

libs/langchain/ragstack_langchain/graph_store/extractors/gliner_link_extractor.py
```diff
@@ -1,56 +1,9 @@
...
-    ) -> Iterable[Set[Link]]:
-        strs = [i if isinstance(i, str) else i.page_content for i in inputs]
-        for entities in self._model.batch_predict_entities(
-            strs, self._labels, **self._extract_kwargs
-        ):
-            yield {
-                Link.bidir(kind=f"{self._kind}:{e['label']}", tag=e["text"])
-                for e in entities
-            }
+from langchain_community.graph_vectorstores.extractors import (
```
bjchambers, line 1: I'm somewhat inclined to delete these sooner rather than later -- we don't want to encourage new dependencies on these. If we keep them around, it should print a warning if you use these rather than the `langchain` provided import paths.

libs/langchain/tests/unit_tests/test_keybert_link_extractor.py
```diff
@@ -1,5 +1,5 @@
-from ragstack_langchain.graph_store.extractors import KeybertLinkExtractor
-from ragstack_langchain.graph_store.links import Link
+from langchain_community.graph_vectorstores.extractors import KeybertLinkExtractor
```
bjchambers, line 1: We can probably drop most/all of these tests:

1. Graph Store functionality is tested in #596.
2. Extractor functionality is tested in langchain.

cbornet: LangChain only runs the unit tests, not the integration ones.
So I removed the ones that are tested in LC's unit tests but kept the ones in integration tests.

libs/langchain/ragstack_langchain/graph_store/extractors/gliner_link_extractor.py
cbornet: Do we know that noone uses them yet ?

PR #586, opened 2024-07-11
feat: Reduce sequential queries for MMR

bjchambers:
This denormalizes the outgoing tags into the `links` table, so that
traversal doesn't need to fetch the set of outgoing tags in an initial
query. This means that traversing is one batch of concurrent queries,
rather than two steps in sequence.

This lowers average retrieval time in 6 experiments from 1.3364s to
1.0920s (would increase as more nodes are retrieved). Writing the
additional copies of the denormalized tags adds a little to indexing
time.


epinzur (approved):
the CQL looks good. I don't have a good enough grasp of the code to comment on the rest of the implementation.

PR #588, opened 2024-07-12
feat: Eliminate the targets table

bjchambers:
This seems to be unnecessary. An SAI index supporting contains seems to be sufficient. Eliminating it removes the need to create the denormalized copy of the data, speeding up insertion and reducing storage needs.


epinzur (approved):

libs/langchain/ragstack_langchain/graph_store/cassandra.py
```diff
@@ -39,7 +39,6 @@ def __init__(
         embedding: Embeddings,
         *,
         node_table: str = "graph_nodes",
-        targets_table: str = "graph_targets",
```
cbornet, line 42: This is a breaking change 😬

PR #590, opened 2024-07-17
Add metadata filtering support to the GraphStore

epinzur:


libs/knowledge-store/ragstack_knowledge_store/graph_store.py
```diff
@@ -243,19 +250,24 @@ def __init__(
...
+            CREATE TABLE IF NOT EXISTS {self.table_name()} (
                 content_id TEXT,
                 kind TEXT,
                 text_content TEXT,
                 text_embedding VECTOR<FLOAT, {embedding_dim}>,
 
                 link_to_tags SET<TUPLE<TEXT, TEXT>>,
                 link_from_tags SET<TUPLE<TEXT, TEXT>>,
-                metadata_blob TEXT,
+                attributes_blob TEXT,
```
epinzur, line 269: changed to the naming used in cassio

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
```diff
@@ -337,7 +381,7 @@ def _nodes_with_ids(
         results: Dict[str, Optional[Node]] = {}
         with self._concurrent_queries() as cq:
 
-            def add_nodes(rows: Iterable[Any]) -> None:
+            def node_callback(rows: Iterable[Any]) -> None:
```
epinzur, line 384: rename to avoid conflict with `add_nodes()` class method.

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
```diff
@@ -90,12 +121,18 @@ def _deserialize_links(json_blob: Optional[str]) -> Set[Link]:
...
+    attributes_blob = row.attributes_blob
+    attributes_dict = (
+        _deserialize_metadata(attributes_blob) if attributes_blob is not None else {}
+    )
     links = _deserialize_links(row.links_blob)
     return Node(
         id=row.content_id,
         text=row.text_content,
-        metadata=metadata,
+        metadata={**attributes_dict, **metadata_s},
```
bjchambers, line 135: I don't know that I'd combine the attributes and the metadata. It seems like for reconstructing the node we should *just* deserialize the attributes blob. When writing, we should put everything there (using that as the source of truth for round tripping), and then treat the `metadata_s` as just an index that supports queries.

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
```diff
@@ -90,12 +121,18 @@ def _deserialize_links(json_blob: Optional[str]) -> Set[Link]:
 
 
 def _row_to_node(row: Any) -> Node:
-    metadata = _deserialize_metadata(row.metadata_blob)
+    metadata_s = row.metadata_s
```
bjchambers, line 124: I don't think this should read metadata_s to create the node. Also, we may want to support multiple versions of this:

`map<str, str>` supports basic equality on strings, integers (their string representation, etc.). But I could see `map<str, int>` to support comparisons on the number, or `map<str, set<str>>` to support `contains` on the fields, etc.

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
bjchambers: I don't love mixing the `attributes` and `metadata` prefixes. It seems like this is *just* writing the metadata a few different ways...

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
```diff
@@ -265,19 +278,47 @@ def _apply_schema(self) -> None:
...
+        self._session.execute(f"""
+            CREATE CUSTOM INDEX IF NOT EXISTS {self._node_table}_metadata_index
+            ON {self.table_name()}(ENTRIES(metadata_s))
             USING 'StorageAttachedIndex';
         """)
 
     def _concurrent_queries(self) -> ConcurrentQueries:
         return ConcurrentQueries(self._session)
 
+    def _parse_metadata(
```
bjchambers, line 300: The term "parse" here feels weird. Any details on what this is actually doing? Maybe some comments and/or a better method name would help.

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
```diff
@@ -670,3 +751,113 @@ def add_targets(rows: Iterable[Any]) -> None:
...
+
+        select_vals = tuple(list(get_cql_vals) + order_cql_vals + limit_cql_vals)
+        select_cql = SELECT_CQL_TEMPLATE.format(
+            columns=columns,
+            table_name=self.table_name(),
+            where_clause=where_clause,
+            order_clause=order_clause,
+            limit_clause=limit_clause,
+        )
+        prepared_query = self._session.prepare(select_cql)
```
bjchambers, line 860: We may need to cache these if the preparation takes too long. Specifically, this will re-preprae the same query multiple times.

epinzur: is it okay to save this for another PR?

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
epinzur: reverted to old name

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
epinzur: reverted to keeping all the metadata pickled in the `metadata_blob` and duplicating indexed fields in `metadata_s`

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
epinzur: fixed

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
epinzur: function removed via other simplifications.

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
epinzur: added to this PR.

bjchambers (approved):

PR #591, opened 2024-07-19
disabled langchain and llamaindex dev tests

epinzur:


nicoloboschi (approved):

cbornet:
Maybe we could have them run on scheduled job only and not on PRs ?

PR #604, opened 2024-07-23
Deactivate LangSmith

cbornet:


scripts/ci-common-env.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
 
-export LANGCHAIN_TRACING_V2=true
+#export LANGCHAIN_TRACING_V2=true
 export LANGCHAIN_ENDPOINT="https://api.smith.langchain.com"
 export LANGCHAIN_PROJECT="ragstack-ci"
```
epinzur, line 5: should we comment out all of these, and add a reason?

# disabling langsmith tracing

epinzur (approved):

scripts/ci-common-env.sh
cbornet: No need to comment everything.
We can add a comment.

PR #609, opened 2024-07-23
Remove ragstack-langchain graph store

cbornet:


epinzur (approved):

bjchambers (approved):

libs/e2e-tests/e2e_tests/langchain/test_keybert_link_extractor.py
bjchambers, line 1: Looks like these are failing because we don't install keybert/gliner in the e2e tests. May need to update things. Also won't need those in `langchain` `pyproject.toml` anymore.

This raises a question -- do we have an e2e test that makes sure that latest release of langchain works with the unreleased version of `knowledge-store`? Would probably be good to verify that upcoming releases won't break things (eg., like the `target_table` removal would have).

cbornet: Yes we test with the main branch of LangChain

cbornet: But we currently don’t have a test with LangChain’s GraphVectorStore.

cbornet: > Yes we test with the main branch of LangChain

Actually those tests \have been deactivated recently: https://github.com/datastax/ragstack-ai/pull/591

PR #612, opened 2024-07-24
[breaking] Fix ColbertVectorStore from_texts

cbornet:


libs/langchain/ragstack_langchain/colbert/colbert_vector_store.py
```diff
@@ -253,22 +255,23 @@ async def afrom_documents(
...
         *,
         database: Optional[ColbertBaseDatabase] = None,
         **kwargs: Any,
     ) -> CVS:
-        if isinstance(embedding, Embeddings):
+        if not isinstance(embedding, TokensEmbeddings):
             raise TypeError(
-                "ColbertVectorStore needs a ColbertBaseEmbeddingModel embedding, "
-                "not an Embeddings object."
+                "ColbertVectorStore requires a ColbertEmbeddings embedding."
```
epinzur, line 266: change to `"ColbertVectorStore requires a TokensEmbeddings embedding."`?

epinzur (approved):

libs/langchain/ragstack_langchain/colbert/embedding.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+class TokensEmbeddings(Embeddings):
+    """Adapter for token-based embedding models and the LangChain Embeddings."""
+
+    def __init__(self, embedding: BaseEmbeddingModel = None):
+        self.embedding = embedding or ColbertEmbeddingModel()
+
+    @override
+    def embed_documents(self, texts: List[str]) -> List[List[float]]:
```
bjchambers, line 16: This doesn't seem right. It changes the signatures of the `Embeddings` class in an incompatible way. I think if we want to go this direction, we should instead have:

1. `TokensEmbeddings` is an interface/ABC like this, but does not extend `Embeddings`.
2. We have a sub-class of `Embeddings` that defines `get_tokens_embeddings() -> TokensEmbeddings` (or just rely on the method being defined).
3. So then subclasses are valid langchain embeddings, and can choose to implement `get_tokens_embeddings()` to get the different interface.

cbornet: I don't get it. In its current form `TokensEmbeddings` is a valid `Embeddings` subclass.
I'm not sure there is a lot of value in introducing yet another interface.
Here `TokensEmbeddings` is a simple wrapper around `BaseEmbeddingModel` that implements `Embeddings` to please the `VectorStore` API.

bjchambers: Ah, yeah. I misread this. It does match the interface.

libs/langchain/ragstack_langchain/colbert/colbert_vector_store.py
cbornet: good catch!

PR #613, opened 2024-07-24
ref: Only prepare once during MMR traversal

bjchambers:


libs/knowledge-store/ragstack_knowledge_store/graph_store.py
```diff
@@ -414,11 +414,20 @@ def mmr_traversal_search(
...
+        columns = "content_id, text_embedding, link_to_tags"
         initial_candidates_query = self._get_search_cql(
             has_limit=True,
-            columns="content_id, text_embedding, link_to_tags",
+            columns=columns,
+            metadata_keys=list(metadata_filter.keys()),
+            has_embedding=True,
+        )
+
+        adjacent_query = self._get_search_cql(
```
epinzur, line 425: note that `_get_search_cql()` has an internal prepare cache now.

Despite that, I think this cleanup makes sense.

epinzur (approved):

PR #634, opened 2024-07-31
feat: mmr traversal starting in neighborhood

bjchambers:
This adds a `neighborhood` parameter to the mmr_traversal, and starts the search with the best candidates adjacent to the provided `neighborhood`.

### DIFF
--- a/libs/knowledge-store/ragstack_knowledge_store/graph_store.py
+++ b/libs/knowledge-store/ragstack_knowledge_store/graph_store.py
@@ -356,7 +356,7 @@ class GraphStore:
         self,
         query: str,
         *,
-        neighborhood: Sequence[str] | None = None,
+        initial_roots: Sequence[str] = (),
         k: int = 4,
         depth: int = 2,
         fetch_k: int = 100,
@@ -377,10 +377,9 @@ class GraphStore:
 
         Args:
             query: The query string to search for.
-            neighborhood: Optional list of documents to use as the initial neighborhood
-                to search. If provided, the adjacent nodes to the neighborhood will be
-                used as the initial candidates, rather than performing a generic vector
-                search.
+            initial_roots: Optional list of document IDs to use for initializing search.
+                The top `adjacent_k` nodes adjacent to each initial root will be included
+                in the set of initial candidates.
             k: Number of Documents to return. Defaults to 4.
             fetch_k: Number of initial Documents to fetch via similarity.
                 Defaults to 100.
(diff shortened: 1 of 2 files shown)


libs/knowledge-store/ragstack_knowledge_store/graph_store.py
```diff
@@ -384,6 +382,10 @@ def mmr_traversal_search(
 
         Args:
             query: The query string to search for.
+            neighborhood: Optional list of documents to use as the initial neighborhood
```
bjchambers, line 385: Note: Currently, providing a neighborhood bypasses the initial fetch. This means that no new "root" nodes will be fetched. This may be a good thing, but I could also be convinced that we should take the union of the neighborhood-adjacencies and the `fetch-k`, and allow `fetch_k = 0` to indicate "only use the neighborhood)... thoughts?

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
```diff
@@ -683,23 +721,24 @@ def _get_outgoing_tags(
...
-                    self._query_source_tags_by_id, (source_id,), callback=add_sources
+                    self._query_ids_and_link_to_tags_by_id, (source_id,), callback=add_sources
                 )
 
         return tags
 
     def _get_adjacent(
         self,
-        tags: set[tuple[str, str]],
+        tags: Set[Tuple[str, str]],
```
bjchambers, line 737: Fixing type annotations. Upper case types (from `typing`) actually type check, while lower case types (from builtins) ignore the parameters, etc.

Also fixing the `B006` errors since those can cause problems.

bjchambers: Actually, it seems like `set` and `tuple` may support type arguments as of 3.9. So I can probably back out those changes and just keep the fix for `B006`.

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
bjchambers: Also, don't love the name...

libs/knowledge-store/tests/integration_tests/test_graph_store.py
```diff
@@ -195,6 +195,17 @@ def test_mmr_traversal(
     results = gs.mmr_traversal_search("0.0", k=4, metadata_filter={"even": True})
     assert _result_ids(results) == ["v0", "v2"]
 
+    # with neighborhood=[v0], we should start traversal there. this means that
+    # the initial candidates are `v2`,`v3`. `v1` is unreachable and not
+    # included.
+    results = gs.mmr_traversal_search("0.0", k=4, neighborhood=["v0"])
+    assert _result_ids(results) == ["v2", "v3"]
```
kerinin, line 202: What would the code look like to re-use results? Would you need to map the ID out of the returned results set? Would it make sense to accept documents as the result set, and then pull the IDs out inside the implementation, or maybe provide a helper method to translate a set of documents into a set of IDs?

kerinin (approved):

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
cbornet: Yes. [PEP585](https://peps.python.org/pep-0585/) and [PEP604](https://peps.python.org/pep-0604/) are supported (through `from __future__ import annotations` for the latter) and enforced.
The only case where we shouldn't use them and ignore the rule is with pydantic v1.
Pydantic v1 is EOL'd so we must not use it.

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
```diff
@@ -809,15 +844,19 @@ def _coerce_string(value: Any) -> str:
 
     def _extract_where_clause_cql(
         self,
-        metadata_keys: list[str] = [],  # noqa: B006
+        has_id: bool = False,
+        metadata_keys: list[str] | None = None,
```
cbornet, line 848: Could also use `metadata_keys: Sequence[str] = ()` and not have to do the None-check

cbornet (approved):
LGTM.
Only a nit.

libs/knowledge-store/tests/integration_tests/test_graph_store.py
bjchambers: I don't think we should try to over-beautify this yet. We're providing components for people to write code, and they should be able to write the for-comprehension to do `[d.id for d in results]`. I also don't think we want to take documents, because they may want to store just the IDs in state somewhere, etc. We don't want to dictate they nede the whole document.

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
bjchambers: Langchain requires Pydantic v1, so we must use it.

PR #650, opened 2024-09-23
DRAFT: moved link storage to a pure metadata-based method

epinzur:
This makes CassandraGraphStore compatible with existing cassio.

Note:
 - only `traversal_search()` has been updated to use the new link storage.


libs/knowledge-store/ragstack_knowledge_store/graph_store.py
```diff
@@ -482,39 +476,33 @@ def fetch_initial_candidates() -> None:
                 # If the next nodes would not exceed the depth limit, find the
                 # adjacent nodes.
                 #
-                # TODO: For a big performance win, we should track which tags we've
+                # TODO: For a big performance win, we should track which links we've
```
bjchambers, line 479: It seems like this may be a stale comment. Specifically, the `difference_update` on line 487 seems to be doing this.

libs/knowledge-store/ragstack_knowledge_store/graph_store.py
```diff
@@ -892,22 +841,27 @@ def _extract_where_clause_params(
...
+        embedding: list[float] | None = None,
+        outgoing_link: Link | None = None,
+    ) -> tuple[PreparedStatement|SimpleStatement, tuple[Any, ...]]:
+        if outgoing_link is not None:
+            if metadata is None:
+                metadata = {}
+            else:
+                # don't add link search to original metadata dict
+                metadata = metadata.copy()
+                metadata[_metadata_s_link_key(link=outgoing_link)] = _metadata_s_link_value()
```
bjchambers, line 858: How does this work? It *looks* like it is doing equality. But, there may be multiple outgoing links, and we need to find the nodes with one of those as an incoming link. It seems like this is maybe going the wrong direction, and also likely missing the set-equality. Are we sure this works the same?

epinzur: when we `add_nodes()`, we are storing all the incoming_links as keys in the `metadata_s` index. There is an arbitrary, static value stored with each key, so that it can be stored in the MAP<text,text> type.

Here, when we search, we are finding all the chunks that have a matching outgoing-link key.

We are essentially doing a hybrid query on metadata keys... the values do not matter.

bjchambers: Ah. That's the part I missed. Could you add a comment to that effect where we store / query? It also seems like there is then a risk that the generated key collides with the user defined key? Is there a prefix or something we should tell people *not* to use in their metadata?

PR #656, opened 2025-03-06
Adding redirect message to graph RAG notebook

briangodsey:
Adding IMPORTANT banner at the top of astra_support.ipynb that redirects to the new graph-rag project.

### DIFF
--- a/libs/knowledge-store/notebooks/astra_support.ipynb
+++ b/libs/knowledge-store/notebooks/astra_support.ipynb
@@ -4,6 +4,24 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "# IMPORTANT: new `graph-rag` project with improved functionality\n",
+    "\n",
+    "> Since the date of publication, some code in this notebook has been superseded by\n",
+    "> a new, dedicated \n",
+    "> [Graph RAG project](https://datastax.github.io/graph-rag/).\n",
+    "> \n",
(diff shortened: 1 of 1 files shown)


libs/knowledge-store/notebooks/astra_support.ipynb
```diff
@@ -1,5 +1,23 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# IMPORTANT: new `graph-rag` project with improved functionality\n",
+    "\n",
+    "> Since the date of publication, some code in this notebook has been superseded by\n",
+    "> a new, dedicated \n",
+    "> [Graph RAG project](https://datastax.github.io/graph-rag/).\n",
+    "> \n",
+    "> Specifically, the `ragstack-ai-knowledge-graph` and\n",
```
bjchambers, line 13: Probably worth mentioning that GraphVectorStore is deprecated. May be worth submitting that change to langchain too now that we have the migration guide (@kerinin @epinzur) to link to.

Specifcally, that is the "user facing" part of the API (they may not know what `ragstack-ai-knowledge-store` is directly, since it provides building blocks for the `GraphVectorStore` in langchain)

briangodsey: GVS isn't in this code---it's `CassandraGraphStore`---but I'll be sure to note that GVS is deprecated in other code.

